PR #4, opened 2024-11-05
DS-2058: Task List component

erincarrieresontario:
- Added ontario-task component to represent individual tasks with status tracking.
- Added ontario-task-list component to manage and display a list of tasks, tracking completed and total tasks dynamically.
- Implemented language support for task status translations.
- Used @State to track task progress and updated counts on render.
- Updated camel casing for badge types to keep it consistent


packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,221 @@
...
+	@Watch('taskStatus')
+	validateTaskStatus() {
+		const isValidStatus = validateValueAgainstArray(this.taskStatus, TaskStatuses);
+		this.taskStatusState = isValidStatus ? this.taskStatus : this.warnDefaultTaskStatus();
+	}
+
+	/**
+	 * Display a console warning if `taskStatus` is invalid and set to a default value.
+	 */
+	private warnDefaultTaskStatus(): TaskStatus {
```
smorrisods, line 89: This method appears to be returning something, I recommend naming it `warnAndGetDefaultTaskStatus` to indicate what it is doing.

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,221 @@
...
+		if (!this.language) {
+			this.language = validateLanguage(event);
+		}
+	}
+
+	/**
+	 * Listen for header language toggling events to update the language dynamically.
+	 */
+	@Listen('headerLanguageToggled', { target: 'window' })
+	handleHeaderLanguageToggled(event: CustomEvent<Language>) {
```
smorrisods, line 130: Iirc this event returns an object with `old` and `new` properties, not just the `Language`.

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,221 @@
...
+	 */
+	private getBadgeColor(): 'black' | 'light-teal' | 'grey' {
+		return BadgeColors[this.taskStatusState];
+	}
+
+	/**
+	 * Provides a translated task status string based on the current language.
+	 */
+	private getTranslatedTaskStatus(): string {
+		return translations.taskStatus[this.taskStatusState][this.language === 'fr' ? 'fr' : 'en'];
```
smorrisods, line 145: Do we have a utility function that will resolve the correct language for us so we don't have to have language validation checks everywhere?

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	/**
+	 * Renders the task label and status content.
+	 */
+	private renderTaskContent() {
+		return (
+			<div class="ontario-task__content">
+				<h2 id="task-label" class="ontario-task__label" aria-label={this.label}>
+					{this.label}
```
smorrisods, line 171: Is the `aria-label` attribute needed when it is the same as the content of the heading?

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,221 @@
...
+	 */
+	@Listen('headerLanguageToggled', { target: 'window' })
+	handleHeaderLanguageToggled(event: CustomEvent<Language>) {
+		this.language = validateLanguage(event);
+	}
+
+	/**
+	 * Determines the badge color based on the current `taskStatusState`.
+	 */
+	private getBadgeColor(): 'black' | 'light-teal' | 'grey' {
```
smorrisods, line 137:
```suggestion
	private getBadgeColour(): 'black' | 'light-teal' | 'grey' {
```

As well, we have a TaskBadgeColour type that can be used as the return type here.

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,221 @@
...
+		return (
+			<div class="ontario-task__content">
+				<h2 id="task-label" class="ontario-task__label" aria-label={this.label}>
+					{this.label}
+				</h2>
+				{this.taskStatusState && (
+					<ontario-badge
+						class="ontario-task__badge"
+						role="status"
+						aria-label={`Task status: ${this.getTranslatedTaskStatus()} (${this.taskStatusState})`}
```
smorrisods, line 177: The text _Task status_ is currently just English and should be handled as a translation.

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,221 @@
...
+	 * This is required to provide the name of the task.
+	 */
+	@Prop() label: string;
+
+	/**
+	 * Specifies an optional link associated with the task.
+	 *
+	 * If provided, clicking the task will navigate to this URL.
+	 */
+	@Prop() link?: string;
```
smorrisods, line 30: I wonder if we should call this `href` rather than `link`, thoughts?

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+	/**
+	 * Disables the task link when set to `true`.
+	 *
+	 * Default is `false`, meaning the link will be active if provided.
+	 */
+	@Prop() deactivateLink: boolean = false;
```
smorrisods, line 37: What's the use case for `deactivateLink`?

packages/ontario-design-system-component-library/src/components/ontario-task-list/ontario-task-list.tsx
smorrisods, line 1: Are there any options needed at this level for the task list?

smorrisods (commented):
Looking pretty good, I've added some questions and comments.  One bigger item I'm wondering about is how we handle groups of tasks?

10000201000005C2000003E83946CD5CFFF76DF2

Perhaps I just missed this when looking through the code.

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
erincarrieresontario: I prefer link 🤔

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
erincarrieresontario: This is if a user cannot yet start the task if they haven't completed a previous task.

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
erincarrieresontario: Not needed, you're right!

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
erincarrieresontario: Or perhaps if the status is completed. It just gives them options.

packages/ontario-design-system-component-library/src/components/ontario-task-list/ontario-task-list.tsx
erincarrieresontario: Options have been added :)

packages/ontario-design-system-component-library-nextjs/README.md
smorrisods, line 1: I will quickly jump on this, we don't need to create a new NextJS package, we just need to update our React package to support SSR.

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,230 @@
...
+	 * Defines the status of the task, with default set to 'NotStarted'.
+	 *
+	 * Accepts values from `TaskStatuses` enum: `NotStarted`, `InProgress`, `Completed`.
+	 */
+	@Prop() taskStatus: TaskStatus;
+
+	/**
+	 * Used for the `aria-describedby` value of the task's label. This will match with the id of the hint text.
+	 */
+	@State() hintTextId: string | null | undefined;
```
smorrisods, line 63: I think this should be marked `private` as well

packages/ontario-design-system-component-library/src/components/ontario-task-list/ontario-task-list.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+	/**
+	 * The number of completed tasks.
+	 */
+	@Prop() completedTasks: number;
+
+	/**
+	 * The total number of tasks.
+	 */
+	@Prop() totalTasks: number;
```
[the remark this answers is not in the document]
smorrisods: I looked into this a bit and it appears there is a way.  Here's the just of it:

We can get all the children of the component and grab data from them with some code.  Using the `componentDidRender` live cycle method should let us get them once their rendered as well.

The collection component:
```tsx
import { Component, h, Element } from '@stencil/core';

@Component({
  tag: 'my-item-collection',
  shadow: true,
})
export class MyItemCollection {
  @Element() el: HTMLElement;

  componentDidRender() {
    const items = this.el.querySelectorAll('my-item');
    items.forEach(item => {
      console.log(item.getAttribute('status'));
    });
  }

  render() {
    return <slot></slot>;
  }
}
```

The child component:
```tsx
import { Component, Prop, h } from '@stencil/core';

@Component({
  tag: 'my-item',
  shadow: true,
})
export class MyItem {
  @Prop() status: string;

  render() {
    return <div>{this.status}</div>;
  }
}
```

I have a few more suggestions from Copilot, but I think this is enough to start noodling this idea. Let me know what you think.

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
matt-stjean: I agree with Scott re: href.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#href

If this prop is just the url, then href is most appropriate. This is the "linked URL".

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
matt-stjean: Negatives can be confusing and the use of a verb makes it sound like an action. 

```suggestion
	 *
	 * Default is `true`, meaning the link will be active if provided.
	 */
	@Prop() active: boolean = true;
```

matt-stjean:
When we make these PRs, can we link out to acceptance criteria or include it? It would be helpful to have screens/figmas/desc when reviewing.

packages/ontario-design-system-component-library/src/components/ontario-hint-text/ontario-hint-text.tsx
```diff
@@ -133,7 +133,7 @@ export class OntarioHintText implements Hint {
 
 	render() {
 		return this.hintContentType === 'string' ? (
-			<p id={this.getId()} class="ontario-hint">
+			<p part="hint-text" id={this.getId()} class="ontario-hint">
```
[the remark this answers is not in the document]
matt-stjean: @smorrisods `part` is part of the Shadow DOM API to expose elements inside a shadow DOM to be styled from outside the shadow DOM using CSS. Normally styles defined outside of Shadow DOM do not affect elements within it. The `part` attribute allows specific elements within the shadow DOM to be targeted for styling from outside the shadow boundary.

In this case the `<p>` element is in the shadow DOM and we can apply styles to it from outside using the `::part()` pseudo-element in CSS. This way we can allow customization without breaking encapsulation. 

Without `part` we would not be able to style this `<p>` element from outside the shadow DOM

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task-statuses.ts
matt-stjean, line 1: This is defining some concepts that overlap with `ontario-badge.types.ts`:

TaskToBadgeColour in ontario-task-statuses.ts maps task statuses to badge colors using a plain object.

BadgeColourToClass in ontario-badge.types.ts maps colors to CSS classes using an enum.

And this makes the logic split across two files. When working on badges or tasks, we must understand both files to make changes.

My loose suggestion is to consolidate badge-related logic, including CSS class mapping and status-to-badge mapping in `ontario-badge.types.ts`

```
// Refactored ontario-badge.types.ts
...
export enum BadgeColourToClass {...}

// Retrieve all badge colours as keys
export const BadgeColours = retrieveEnumKeys(BadgeColourToClass);
export type BadgeColour = keyof typeof BadgeColourToClass;

// Map task statuses to badge colours...
export const TaskToBadgeColour: Record<TaskStatus, BadgeColour> = {
notStarted: 'light-teal',
  inProgress: 'teal',
  completed: 'white',
  cannotStartYet: 'dark-grey',
  error: 'red',
  optional: 'grey',
};

// Get CSS class from a task status
export function getBadgeClassFromTaskStatus(status: TaskStatus): string {
  const badgeColour = TaskToBadgeColour[status];
  return BadgeColourToClass[badgeColour];
}
```

This would free up `ontario-task-statuses.ts` to focus on task-specific logic, with badge-related logic moved to `ontario-badge.types.ts`

We could create an enum for task statuses:
```
// Enum for task statuses
export const TaskStatuses = ['notStarted', 'inProgress', 'completed', 'cannotStartYet', 'error', 'optional'] as const;
export type TaskStatus = (typeof TaskStatuses)[number];
```

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	/**
+	 * Determines the badge colour based on the current `taskStatusState`.
+	 *
+	 * @returns {TaskBadgeColour}
+	 */
+	private getBadgeColour(): TaskBadgeColour {
+		return TaskToBadgeColour[this.taskStatusState] || 'grey';
+	}
```
matt-stjean, line 141: With the `ontario-task-statuses.ts` suggestion I made, this would just change to:
```private getBadgeColour(): string {
         return getBadgeClassFromTaskStatus(this.taskStatusState);
}
```

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task-statuses.ts
matt-stjean: Enums and Record types ensure mappings are consistent and type-checked.

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.scss
```diff
@@ -0,0 +1,57 @@
+@use 'sass:math';
+@use '@ongov/ontario-design-system-global-styles/dist/styles/scss/1-variables/spacing.variables' as spacing;
+@use '@ongov/ontario-design-system-global-styles/dist/styles/scss/1-variables/colours.variables' as colours;
+@use '@ongov/ontario-design-system-global-styles/dist/styles/scss/2-tools/functions/global.functions' as globalFunctions;
+@use '@ongov/ontario-design-system-global-styles/dist/styles/scss/1-variables/breakpoints.variables' as breakpoints;
+
+.ontario-task {
+	border-bottom: 2px solid colours.$ontario-greyscale-20;
```
matt-stjean, line 8: Replace hardcoded with appropriate variables. I am not sure if these exist but it's just an example.
```
border-bottom: spacing.$border-width-thick solid colours.$ontario-greyscale-20;
padding: spacing.$spacing-4 spacing.$spacing-0;
```

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.scss
```diff
@@ -0,0 +1,57 @@
...
+	padding: globalFunctions.px-to-rem(19) spacing.$spacing-0;
+
+	&__content {
+		display: flex;
+		justify-content: space-between;
+		align-items: center;
+
+		@media screen and (max-width: breakpoints.$small-breakpoint) {
+			display: block;
+		}
```
matt-stjean, line 18: `@media` queries nested within individual selectors makes it hard to manage breakpoints and scale. Can we centralize the media queries, grouped at the end of the file or in a responsive section of the file?

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.scss
```diff
@@ -0,0 +1,57 @@
...
+	&__link {
+		display: block;
+		text-decoration: underline;
+		color: colours.$ontario-colour-link;
+		font-weight: normal;
+	}
+
+	&__hint-text--true {
+		margin-bottom: spacing.$spacing-2;
+	}
```
matt-stjean, line 48: Is this modifier class meant for when the hint-text is visible? Could we use something more descriptive like `&__hint-text--visible`

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.scss
```diff
@@ -0,0 +1,57 @@
...
+			}
+		}
+	}
+
+	&__link {
+		display: block;
+		text-decoration: underline;
+		color: colours.$ontario-colour-link;
+		font-weight: normal;
+	}
```
matt-stjean, line 44: Do we need distinct focus and hover states defined for these links? like
```
&:hover,
  &:focus {
    text-decoration: none;
    color: colours.$ontario-colour-link-hover; // Or whatever it actually is
  }
  ```

File: packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.scss [hunk 17]
```diff
@@ -0,0 +1,57 @@
...
+}
+
+ontario-hint-text::part(hint-text) {
+	margin: spacing.$spacing-0;
+
+	@media screen and (max-width: breakpoints.$small-breakpoint) {
+		margin-top: spacing.$spacing-2;
+	}
+}
```
matt-stjean, line 57: Is styling of `ontario-hint-text::part(hint-text)` here safe? Will it impact other components using `ontario-hint-text`?

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,230 @@
...
+	 * The hint text options are re-assigned to the internalHintText array.
+	 */
+	@State() private internalHintText: Hint;
+
+	/**
+	 * Mutable variable, for internal use only.
+	 *
+	 * Set the task's status state depending on validation result.
+	 */
+	@State() private taskStatusState: TaskStatus;
```
matt-stjean, line 75: This duplicates `taskStatus` prop, we should remove this `taskStatusState` and validate the prop directly when needed...We can replace `taskStatusState` with:
```
private getValidatedTaskStatus(): TaskStatus {
  const isValidStatus = validateValueAgainstArray(this.taskStatus, TaskStatuses);
  return isValidStatus ? this.taskStatus : this.warnAndGetDefaultTaskStatus();
}
```

And then anywhere taskStatusState is used, we can call `this.getValidatedTaskStatus()`

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,230 @@
...
+					</ontario-badge>
+				)}
+			</div>
+		);
+	}
+
+	async componentDidLoad() {
+		this.hintTextId = await this.hintTextRef?.getHintTextId();
+	}
+
```
matt-stjean, line 199: This could lead to a race condition. This should be set synchronously during `componentWillLoad` if that is possible. Otherwise, if it must stay async, we should provide a fallback or something.

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,230 @@
...
+					ref={(el) => (this.hintTextRef = el)}
+				></ontario-hint-text>
+			);
+		}
+		return null;
+	}
+
+	private getClass(): string {
+		return this.hintText ? `ontario-task__label ontario-task__hint-text--true` : `ontario-task__label`;
+	}
```
matt-stjean, line 169: Could be made cleaner by using an array and join logic:

```
private getClass(): string {
  return ['ontario-task__label', this.hintText && 'ontario-task__hint-text--true']
      .filter(Boolean)
      .join(' ');
}
```

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,230 @@
...
+				) : (
+					<div>
+						{this.renderTaskContent()}
+						{this.renderHintText()}
+					</div>
+				)}
+			</article>
+		);
+	}
+}
```
matt-stjean, line 230: The logic for rendering link and non-link wrappers is duplicating `renderTaskContent` and `renderHintText`. Can we abstract common content into a variable and use conditional wrapping instead?

So like 
```
const taskContent = (
   <>
       {this.renderTaskContent()}
       {this.renderHintText()}
   </>
);
```
```
<article ...>
    {this.link && !this.deactivateLink ? ( <a...> {taskContent} </a>) 
          : ( <div>{taskContent}</div>
     )}
</article>
```

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,230 @@
...
+		this.language = validateLanguage(event);
+	}
+
+	/**
+	 * Determines the badge colour based on the current `taskStatusState`.
+	 *
+	 * @returns {TaskBadgeColour}
+	 */
+	private getBadgeColour(): TaskBadgeColour {
+		return TaskToBadgeColour[this.taskStatusState] || 'grey';
```
matt-stjean, line 140: With the validated taskStatus and centralized badge logic we can do:
```
return TaskToBadgeColour[this.getValidatedTaskStatus()] || 'grey';
```

packages/ontario-design-system-component-library/src/components/ontario-task/test/ontario-task.e2e.ts
```diff
@@ -0,0 +1,11 @@
...
+
+describe('ontario-task', () => {
+	it('renders', async () => {
+		const page = await newE2EPage();
+		await page.setContent('<ontario-task></ontario-task>');
+
+		const element = await page.find('ontario-task');
+		expect(element).toHaveClass('hydrated');
+	});
+});
```
matt-stjean, line 11: Similar comments to the ontario-task-list test files.

smorrisods:
> When we make these PRs, can we link out to acceptance criteria or include it? It would be helpful to have screens/figmas/desc when reviewing.

There aren't any for this one due to it being developed before the specifications have been created.  It isn't in a state of being ready to go, mostly in a state of exploration.  We'll want to figure out how to handle this case, @matt-stjean, where the tech exploration of the design is handled should we need to develop something to get an idea how it works.

matt-stjean:
> > When we make these PRs, can we link out to acceptance criteria or include it? It would be helpful to have screens/figmas/desc when reviewing.
> 
> There aren't any for this one due to it being developed before the specifications have been created. It isn't in a state of being ready to go, mostly in a state of exploration. We'll want to figure out how to handle this case, @matt-stjean, where the tech exploration of the design is handled should we need to develop something to get an idea how it works.

Thanks for clarifying, I understand!

packages/ontario-design-system-component-library/src/components/ontario-hint-text/ontario-hint-text.tsx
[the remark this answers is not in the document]
erincarrieresontario: Thanks @matt-stjean :)

packages/ontario-design-system-component-library/src/components/ontario-task-list/ontario-task-list.tsx
```diff
@@ -0,0 +1,79 @@
...
+		this.totalTasks = tasks.length;
+
+		const resolvedLanguage = validateLanguage(this.language);
+
+		const completedTranslation =
+			translations.taskStatus.completed[resolvedLanguage as keyof typeof translations.taskStatus.completed];
+
+		this.completedTasks = Array.from(tasks).filter(
+			(task) => task.getAttribute('data-task-status') === completedTranslation,
+		).length;
```
smorrisods, line 50: We should be able to use the `taskStatus` property to check this rather than comparing through translations.

smorrisods (changes requested):
I've added a few comments and questions.  Things are looking really good and this is mostly cleanup for some minor things.

packages/ontario-design-system-component-library/src/components/ontario-badge/ontario-badge.scss
```diff
@@ -69,3 +69,14 @@
...
+.ontario-badge--white {
+	background-color: colours.$ontario-colour-white;
+	color: colours.$ontario-greyscale-70;
+	border: globalFunctions.px-to-rem(1) solid colours.$ontario-greyscale-70;
+}
+
+.ontario-badge--grey {
+	background-color: colours.$ontario-greyscale-20;
+	color: colours.$ontario-colour-black;
+}
```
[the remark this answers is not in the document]
erincarrieresontario: I think thats a good idea - but should be tackled in a separate ticket!

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	/**
+	 * This listens for the `headerLanguageToggled` event sent from the language toggle when it is is connected to the DOM.
+	 * It is used for changing the component language after the language toggle has been activated.
+	 */
+	@Listen('headerLanguageToggled', { target: 'window' })
+	handleHeaderLanguageToggled(event: CustomEvent<Language>) {
+		this.language = validateLanguage(event);
+	}
```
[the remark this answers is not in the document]
erincarrieresontario: This is following the same pattern as all of the other components...is this something we want to change across the board @smorrisods ? If so, we should make a ticket for this.

packages/ontario-design-system-component-library/src/components/ontario-task-list/ontario-task-list.tsx
```diff
@@ -0,0 +1,79 @@
...
+	/**
+	 * The label prop used for the task list heading.
+	 */
+	@Prop() label: string;
+
+	/**
+	 * The language of the component.
+	 * This is used for translations, and is by default set through event listeners checking for a language property from the header. If none are passed, it will default to English.
+	 */
+	@Prop({ mutable: true }) language?: Language = 'en';
```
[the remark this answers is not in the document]
smorrisods: @kyle-binger-3 bumping this for your review

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
smorrisods: The reason we use the state is to hold the value internally.  The task status is the external facing member and the the state is the internal state, which has already been validated and is kept up-to-date should the property change.

Changes to State also trigger a re-render should they change, so this wires us into the life cycle model of the component.

https://stenciljs.com/docs/state#state

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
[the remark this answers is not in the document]
smorrisods: You're correct, this is used across all components, except none of them match the actual definition of the event.

```ts
/**
* Event that fires when the language toggle is pressed/clicked.
*
* The event contains the oldLanguage along with the newLanguage.
*/
@Event() headerLanguageToggled: EventEmitter<HeaderLanguageToggleEventDetails>;
```

They all need to be updated, but new components shouldn't follow the other components if their definitions are incorrect.  I will make an issue to update the others, but this one can be fixed here.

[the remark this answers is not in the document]
smorrisods: See DS-2154 for the follow up

packages/ontario-design-system-component-library/src/components/ontario-task-list/ontario-task-list.tsx
```diff
@@ -0,0 +1,79 @@
...
+		const tasks = this.el.querySelectorAll('ontario-task');
+		this.totalTasks = tasks.length;
+
+		const resolvedLanguage = validateLanguage(this.language);
+
+		const completedTranslation =
+			translations.taskStatus.completed[resolvedLanguage as keyof typeof translations.taskStatus.completed];
+
+		this.completedTasks = Array.from(tasks).filter(
+			(task) => task.getAttribute('data-task-status') === completedTranslation,
```
smorrisods, line 49: This looks like it is doing string comparisons rather than using the task status enum from the task elements.  We shouldn't have to rely on the translations to find out that status, this is fragile and loosely couples the task status with the task list

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,254 @@
...
+						{this.getTranslatedTaskStatus()}
+					</ontario-badge>
+				)}
+			</div>
+		);
+	}
+
+	async componentDidLoad() {
+		const hintId = await this.hintTextRef?.getHintTextId();
+		this.hintTextId = hintId || 'default-hint-id'; // Fallback to a default value
```
smorrisods, line 214: `default-hint-id` is used in multiple places but isn't a constant.  I recommend making it a constant and naming it so we have a definition of it instead of it being a magic string.

MeaghanCarrieres:
Screenshot 2025-02-13 at 4 51 55 PM

Hey @erincarrieresontario, it looks like there may be missing padding on the sides of the task list items, or is this the intended style?

packages/ontario-design-system-component-library/src/components/ontario-task-list/ontario-task-list.tsx
```diff
@@ -0,0 +1,79 @@
...
+		this.countTasks();
+	}
+
+	render() {
+		// Resolve the language to ensure valid translations are used.
+		const resolvedLanguage = validateLanguage(this.language);
+
+		return (
+			<div class="ontario-task-list__container">
+				<h2 class="ontario-task-list__heading">{this.label}</h2>
```
MeaghanCarrieres, line 63: Will this always be an H2, and will a user always have a label for the tasks? I notice in your example in `index.html`, there is no label passed but an H2 is still generated. Perhaps we want to either validate that a user has passed in a label, or if not, make it optional and not coded in if no label prop is present.

packages/ontario-design-system-component-library/src/components/ontario-task-list/ontario-task-list.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+				<p class="ontario-task-list__completion-text" aria-live="polite">
```
MeaghanCarrieres, line 65: I'm curious about the `aria-live` property here. Is the intention that this content will be updating often?

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,254 @@
...
+	 * Defines the status of the task, with default set to 'NotStarted'.
+	 *
+	 * Accepts values from `TaskStatuses` enum: `NotStarted`, `InProgress`, `Completed`.
+	 */
+	@Prop() taskStatus: TaskStatus;
+
+	/**
+	 * Used for the `aria-describedby` value of the task's label. This will match with the id of the hint text.
+	 */
+	@State() private hintTextId: string | null | undefined;
```
MeaghanCarrieres, line 72: I wonder if we should do something similar to the checkboxes/radio buttons here where an ID can be generated to match the aria-describedBy value

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+		const taskContent = (
+			<Fragment>
+				{this.renderTaskContent()}
+				{this.renderHintText()}
+			</Fragment>
+		);
+
+		return (
+			<article
```
MeaghanCarrieres, line 238: Hey @erincarrieresontario, was there a reason you chose the `article` tag vs. an actual list? Just wondering what makes the most sense semantically.

smorrisods: I was just looking into this and it seems maybe it has something to do with 🤔 

> The <article> [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML) element represents a self-contained composition in a document, page, application, or site, which is intended to be independently distributable or reusable (e.g., in syndication). Examples include: a forum post, a magazine or newspaper article, or a blog entry, a product card, a user-submitted comment, an interactive widget or gadget, **_or any other independent item of content_**.

[Source](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article)

packages/ontario-design-system-component-library/src/components/ontario-task-list/ontario-task-list.tsx
erincarrieresontario: 👍

erincarrieresontario:
> Screenshot 2025-02-13 at 4 51 55 PM
> 
> Hey @erincarrieresontario, it looks like there may be missing padding on the sides of the task list items, or is this the intended style?

Hey @MeaghanCarrieres - this is the indented style as shown in the Figma design :)

packages/ontario-design-system-component-library/src/components/ontario-task-list/ontario-task-list.tsx
erincarrieresontario: There should always be a label though, it is a required prop. I will give the user the ability to pass in the heading level!

packages/ontario-design-system-component-library/src/components/ontario-task-list/ontario-task-list.tsx
erincarrieresontario: The paragraph displays the current task completion status, which can change as tasks are completed or added. By setting aria-live="polite", ythe like screen readers will announce the updated information to the user without interrupting their current activity.

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
erincarrieresontario: Using an <article> is semantically appropriate if you consider each task to be a self-contained, independent piece of content. An <article> is meant for content that could stand on its own and be redistributed or repurposed separately. In this case, if each task is seen as an individual unit (especially if it includes its own heading, link, and status information), using an <article> is justified. Furthermore, since the tasks can be tightly integrated as items within a larger list or sequence, the task-list component has each task within an <li> element!

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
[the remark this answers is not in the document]
erincarrieresontario: Fixed!

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
erincarrieresontario: Fixed!

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
erincarrieresontario: updated!

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
MeaghanCarrieres: Hey @erincarrieresontario, just reading this last part of your comment here:
`the task-list component has each task within an li element` - is the intended structure to be article, and then unordered list with list items?

If so, that isn't what i'm seeing is being generated in the component library index file. It seems like articles with heading tags and paragraphs. Should it be lists instead?

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,282 @@
...
+@Component({
+	tag: 'ontario-task',
+	styleUrl: 'ontario-task.scss',
+	shadow: true,
+})
+export class OntarioTask {
+	@Element() el: HTMLElement;
+
+	// Reference to the ontario-hint-text element for this task.
+	hintTextRef?: HTMLOntarioHintTextElement;
```
MeaghanCarrieres, line 20: Could this potentially be undefined?

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,282 @@
...
+	 * This is used for translations, and is by default set through event listeners checking for a language property from the header. If no language is passed, it will default to English.
+	 */
+	@Prop({ mutable: true }) language?: Language;
+
+	/**
+	 * Used to include the ontario-hint-text component for the task.
+	 *
+	 * This is optional.
+	 */
+	@Prop() hintText?: string | Hint;
```
MeaghanCarrieres, line 62: Should we set this prop as mutable?

MeaghanCarrieres:
Hey @erincarrieresontario, just noticed the component looks a bit funny on mobile:

Screenshot 2025-02-20 at 1 15 39 PM

packages/ontario-design-system-component-library/src/components/ontario-task-list/ontario-task-list.tsx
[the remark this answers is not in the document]
kyle-binger-3: I think this setup is okay @smorrisods, outside of the header and language toggle most of our components that need language still utilize it as a prop.

One thing (in the future) we may want to think about standardizing, is whether content we pass in to components  should be wrapped in a language object, e.g. "en": {}, "fr": {}

or if a component should even care about that, and just have the information change as needed when passed to our components. I've seen both kinds of implementations across our components project.

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
erincarrieresontario: Updating a check for this!

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
erincarrieresontario: Updated so that each task is now wrapped in an li!

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
erincarrieresontario: yes!

packages/ontario-design-system-component-library/src/components/ontario-hint-text/readme.md
```diff
@@ -59,6 +59,12 @@ Type: `Promise<string | undefined>`
 
 Promise<string | undefined>
 
+## Shadow Parts
+
+| Part          | Description |
+| ------------- | ----------- |
+| `"hint-text"` |             |
+
```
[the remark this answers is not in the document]
erincarrieresontario: I can't seem to add a description for this - it gets removed. Not sure where it is being populated.

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,314 @@
...
+	 */
+	private renderTaskContent() {
+		const headingProps: any = {
+			id: `task-label--${this.taskId}`,
+			class: this.getClass(),
+		};
+
+		return (
+			<div class="ontario-task__content">
+				{h(this.headingLevel, headingProps, this.label)}
```
MeaghanCarrieres, line 240: Is this `h` at the beginning a typo?

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,314 @@
...
+		const taskContent = (
+			<Fragment>
+				{this.renderTaskContent()}
+				{this.renderHintText()}
+			</Fragment>
+		);
+
+		return (
+			<li class="ontario-task-list__task">
+				<article
```
MeaghanCarrieres, line 296: I'm still kind of curious what the point of using an article is here if we are using list items?

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
erincarrieresontario: Yes 👍

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
erincarrieresontario: You're right - I think because we have the <li> embedded now, we can remove the article.

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
erincarrieresontario: its not!

packages/ontario-design-system-component-library/src/components/ontario-hint-text/readme.md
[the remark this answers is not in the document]
MeaghanCarrieres: I did a bit of googling, and you might be able to do something like:

```
/**
* @part hint-text - description here
*/
```

in the `ontario-hint-text` component to populate it.

MeaghanCarrieres (approved):

packages/ontario-design-system-component-library/src/components/ontario-task/ontario-task.tsx
```diff
@@ -0,0 +1,312 @@
...
+		);
+	}
+
+	/**
+	 * After the component loads, retrieve the hint text ID (if available) from the hintText component,
+	 * and set it for the `aria-describedby` attribute.
+	 */
+	async componentDidLoad() {
+		this.el.setAttribute('data-task-status', this.taskStatus);
+	}
```
kyle-binger-3, line 263: This seems to be setting `data-task-status` and not `aria-describedby`

kyle-binger-3 (changes requested):
Hey @erincarrieresontario! Overall it looks awesome! I left some minor feedback and notes.

packages/ontario-design-system-component-library-react/package.json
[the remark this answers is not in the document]
erincarrieresontario: oh, good catch, this was added just to run the tests, but it actually is not needed!

kyle-binger-3 (approved):
Looks great @erincarrieresontario!

packages/ontario-design-system-component-library/src/components/ontario-hint-text/readme.md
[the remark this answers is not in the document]
erincarrieresontario: This is not working - I am opening up a new ticket so this does not hold up the entire MR: https://ontariodigital.atlassian.net/browse/DS-2185

MeaghanCarrieres:
Look good to me, great work @erincarrieresontario !

MeaghanCarrieres (approved):

packages/ontario-design-system-component-library/src/components/ontario-hint-text/test/ontario-hint-text.spec.tsx
```diff
@@ -24,9 +24,9 @@ describe('ontario-hint-text', () => {
 		expect(page.root).toEqualHtml(`
       <ontario-hint-text element-id="idTest" hint="This is a test hint">
         <mock:shadow-root>
-          <p id="idTest" class="ontario-hint">
-            This is a test hint
-          </p>
+					<p id="idTest" class="ontario-hint" part="hint-text">
+						This is a test hint
+					</p>
```
matt-stjean, line 29: should this be so indented?

packages/ontario-design-system-component-library/src/components/ontario-task-list/readme.md
```diff
@@ -0,0 +1,53 @@
...
+
+Use a task list to show the user activities they have completed and what they have left to do.
+
+## Usage guidance
+
+Please refer to the [Ontario Design System](https://designsystem.ontario.ca/components/detail/task-list.html) for current documentation guidance for tables.
+
+## Configuration
+
+Once the component package has been installed (see Ontario Design System Component Library for installation instructions), the task component can be added directly into the project's code, and can be customized by updating the properties outlined [here](#properties). Additional information on custom types for header properties are outlined [here](#custom-property-types). Please see the [examples](#examples) below for how to configure the component.
```
matt-stjean, line 13: Maybe we could make a small pre-req section instead of preambling the configuration with a requirement?

packages/ontario-design-system-component-library/src/components/ontario-hint-text/test/ontario-hint-text.spec.tsx
smorrisods: Looks like the file is a mix of tabs and spaces, hence the inconsistent formatting.

packages/ontario-design-system-component-library/src/components/ontario-task-list/ontario-task-list.tsx
```diff
@@ -0,0 +1,166 @@
...
+	 * The label prop used for the task list heading.
+	 */
+	@Prop() label: string;
+
+	/**
+	 * Allows consumers to define the heading level for the task list component.
+	 *
+	 * Accepts 'h1', 'h2', 'h3' or 'h4'. Default is 'h3'.
+	 */
+	@Prop() headingLevel: LimitedHeadingLevel = 'h2';
```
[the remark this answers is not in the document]
erincarrieresontario: This makes sense, Ive updated 👍

smorrisods (changes requested):

packages/ontario-design-system-component-library/src/components/ontario-hint-text/ontario-hint-text.tsx
```diff
@@ -40,6 +40,10 @@ export class OntarioHintText implements Hint {
 
 	@State() hintState: string;
 
+	/**
+	 * @part hint-text - The container for the hint text content. This part can be used to apply custom styles to the hint text.
```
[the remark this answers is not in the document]
erincarrieresontario: @smorrisods ,yes I had already tried this solution, but it didn't work!

packages/ontario-design-system-component-library/src/components/ontario-task-list/readme.md
```diff
@@ -0,0 +1,73 @@
...
+		></ontario-task>
+		<ontario-task
+			label="Task 2"
+			task-id="Task-2-unique-id"
+			hint-text="A hint for task 2"
+			task-status="inProgress"
+			link="https://example.com"
+			deactivate-link="true"
+		>
+	</ontario-task-list>
```
smorrisods, line 59: For consistency since this is React based, these should use the React components imported at the top:

```suggestion
	<OntarioTaskList language="en">
		<OntarioTask
			label="Task 1"
			task-id="Task-1-unique-id"
			hint-text="A hint for task 1"
			task-status="completed"
			link="https://example.com"
		></OntarioTask>
		<OntarioTask
			label="Task 2"
			task-id="Task-2-unique-id"
			hint-text="A hint for task 2"
			task-status="inProgress"
			link="https://example.com"
			deactivate-link="true"
		></OntarioTask>
	</OntarioTaskList>
```

packages/ontario-design-system-component-library/src/components/ontario-task-list/readme.md
```diff
@@ -0,0 +1,73 @@
+import { OntarioTaskList } from '@ongov/ontario-design-system-component-library-react';
```
smorrisods, line 1: The example is using the `OntarioTask` component as well, so it should be imported:

```suggestion
import { OntarioTaskList, OntarioTask } from '@ongov/ontario-design-system-component-library-react';
```

smorrisods (changes requested):
This is looking great.  Some small changes in the Readme files for the live examples, they're currently using the HTML versions of the components rather than the imported React versions.  Outside those suggestions everything is ready to go.

smorrisods:
I will note, looking at the commit subject our release notes (`CHANGELOG.md`) are going to get spammed with a lot of extra stuff that doesn't relate to the actual Task List and Task components. I recommend cleaning that up before we merge so we keep our release notes as clean as we can.

image

packages/ontario-design-system-component-library/src/components/ontario-hint-text/readme.md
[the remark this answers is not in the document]
smorrisods: I just tried how they did it in the linked issue and it worked!

image

smorrisods (approved):

PR #7, opened 2024-11-26
Alpha -> Develop

smorrisods:
This MR brings the `alpha` branch back into `develop`.  It includes:

- Updates to the CI/CD workflow to enable npm package publishing.


matt-stjean:
@smorrisods The commits do not adhere our commit style (conventional commits).

The commits with "chore(release)" don't match the style either but because "release" isn't a scope. I suggest these ones refer to the scope of the change, so in this case `chore(ontario-design-system): release xxxx` or `chore(monorepo): release xxxx`.

The `ci: xxx` commit should include scope as well, `commitlint` should have warned you and stopped it.

```
<type>[optional scope]: <description>

[optional body]

[optional footer(s)]
```

smorrisods:
@matt-stjean I looked through our past commits and this is the same commit message we've been using from day one.  Each time this commit appears it is related to a release of the npm packages, so I'd argue `release` is a scope.  We do have control over this message, so we can customize it to whatever we like, but historically we've kept it simple.

As a general `scope` observation, the `scope` is _optional_ and in the case of the `ci:` when it impacts more than one file I tend to leave to scope out as `ci:` is highly specific to the CI workflow.

The scope being _optional_  is inline with the sample you shared about the conventional commits and the tooling respects this state - I don't believe we even offer a warning about this.

If you'd like, we can gave a quick call and iron this out.

PR #8, opened 2024-12-09
DS-2103: isNumber function validation inverted

smorrisods:
This fix handles validating numbers are numbers and not numbers are NaN.
- Add unit tests to valid this case via the `date-validation-utils function` `isInvalidYear`.

Closes DS-2103

### DIFF
--- a/packages/ontario-design-system-component-library/src/components/ontario-date-input/test/ontario-date-input.spec.tsx
+++ b/packages/ontario-design-system-component-library/src/components/ontario-date-input/test/ontario-date-input.spec.tsx
@@ -1,5 +1,6 @@
 import { newSpecPage } from '@stencil/core/testing';
 import { OntarioDateInput } from '../ontario-date-input';
+import { isInvalidYear } from '../utils';
 
 describe('ontario-date-input', () => {
 	it('renders deafult state', async () => {
@@ -118,5 +119,80 @@ describe('ontario-date-input', () => {
 				</mock:shadow-root>
 			</ontario-date-input>
 		`);
+	});
+});
+
+describe('date-validation-utils', () => {
+	it('is invalid year value - undefined string', () => {
+		// Note: Type system doesn't like `undefined` as `any` lets us force it for testing purposes
+		const value: any = undefined;
+
+		const isInvalidYearResult = isInvalidYear(value);
+		expect(isInvalidYearResult).toEqual(true);
+	});
+
+	it('is invalid year value - null string', () => {
+		// Note: Type system doesn't like `null` as `any` lets us force it for testing purposes
+		const value: any = null;
+
+		const isInvalidYearResult = isInvalidYear(value);
+		expect(isInvalidYearResult).toEqual(true);
+	});
+
+	it('is invalid year value - empty string', () => {
+		const value = ''; // Empty string
+
+		const isInvalidYearResult = isInvalidYear(value);
+		expect(isInvalidYearResult).toEqual(true);
+	});
+
+	it('is invalid year value - written out number', () => {
+		const value = 'two-thousand';
+
+		const isInvalidYearResult = isInvalidYear(value);
+		expect(isInvalidYearResult).toEqual(true);
+	});
+
+	it('is valid year value', () => {
+		const value = '2000';
+
+		const isValidYear = !isInvalidYear(value);
+		expect(isValidYear).toEqual(true);
+	});
+
+	it('is valid year when in range', () => {
+		const value = '2000';
+		const minYear = 1;
+		const maxYear = 9999;
+
+		const isValidYear = !isInvalidYear(value, minYear, maxYear);
+		expect(isValidYear).toEqual(true);
+	});
+
+	it('is valid year when in range but is minYear', () => {
+		const value = '2000';
+		const minYear = 2000;
+		const maxYear = 9999;
+
+		const isValidYear = !isInvalidYear(value, minYear, maxYear);
+		expect(isValidYear).toEqual(true);
+	});
+
+	it('is valid year when in range but is maxYear', () => {
+		const value = '2000';
+		const minYear = 1;
+		const maxYear = 2000;
+
+		const isValidYear = !isInvalidYear(value, minYear, maxYear);
+		expect(isValidYear).toEqual(true);
+	});
+
+	it('is invalid year when out of range', () => {
+		const value = '2000';
+		const minYear = 1;
+		const maxYear = 1999;
+
+		const isValidYear = !isInvalidYear(value, minYear, maxYear);
+		expect(isValidYear).toEqual(false);
 	});
 });
--- a/packages/ontario-design-system-component-library/src/components/ontario-date-input/utils/date-validation-utils.ts
+++ b/packages/ontario-design-system-component-library/src/components/ontario-date-input/utils/date-validation-utils.ts
@@ -35,17 +35,31 @@ const isInvalidMonth = (value: string) => {
 	return month < MONTH_MIN_VALUE || month > MONTH_MAX_VALUE;
 };
 
-/*
- * Year field should be a number, not negative, at length at least 4 digits
+/**
+ * Valid value year against minimum and maximum year range (defaulted if not provided.)
+ *
+ * Year field should be a number and not written out; it has no concept of positive and negative.
+ *
+ * Suggestion:
+ * - `< 0` B.C.E.
+ * - `>= 0` A.C.E.
+ *
+ * @param value value to check for validity
+ * @param minYear minimum valid year to validate against
+ * @param maxYear maximum valid year to validate against
  */
-const isInvalidYear = (value: string, minYear: number = YEAR_MIN_VALUE, maxYear: number = YEAR_MAX_VALUE) => {
+export const isInvalidYear = (
+	value: string | number,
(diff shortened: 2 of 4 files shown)


packages/ontario-design-system-component-library/src/utils/helper/utils.ts
```diff
@@ -67,7 +67,7 @@ export function removeObjectsBySpecificKey<T>(objects: T[], key: keyof T, value:
  * @returns {boolean}
  */
 export function isNumber(value: string | number): boolean {
-	return isNaN(Number(value)) === true;
+	return isNaN(Number(value)) === false;
```
matt-stjean, line 70: For consistency and clarity, this could be named `isValidNumber`?

packages/ontario-design-system-component-library/src/components/ontario-date-input/utils/date-validation-utils.ts
```diff
@@ -38,7 +38,7 @@ const isInvalidMonth = (value: string) => {
 /*
  * Year field should be a number, not negative, at length at least 4 digits
  */
```
matt-stjean, line 40: Consider extending the comment to explain that isInvalidYear validates numeric and range-bound conditions. The existing comment is more like supplemental text.

packages/ontario-design-system-component-library/src/components/ontario-date-input/test/ontario-date-input.spec.tsx
```diff
@@ -120,3 +121,25 @@ describe('ontario-date-input', () => {
...
+	it('is valid year when in range', () => {
+		const isValidYear = !isInvalidYear('2000', 1, 9999);
+		expect(isValidYear).toEqual(true);
+	});
+
+	it('is invalid year when out of range', () => {
+		const isValidYear = !isInvalidYear('2000', 1, 1999);
+		expect(isValidYear).toEqual(false);
+	});
+});
```
matt-stjean, line 145: Great test coverage but consider adding a couple edge cases

- Empty strings ('') and null/undefined values.
- Boundary conditions (e.g., exactly at minYear and maxYear).

matt-stjean (commented):
Looks good, minor suggestions to test coverage and readability.

packages/ontario-design-system-component-library/src/utils/helper/utils.ts
smorrisods: The goal here is that the `value` is a `number` the validity of the number is up to the caller to figure out.  If it is a number, is it a valid number? This only does the first part of the check.

packages/ontario-design-system-component-library/src/components/ontario-date-input/utils/date-validation-utils.ts
smorrisods: Good catch, I'll update this!

packages/ontario-design-system-component-library/src/components/ontario-date-input/test/ontario-date-input.spec.tsx
smorrisods: Great idea! 😸

erincarrieresontario:
This overall looks really great @smorrisods. I'm wondering if there is a way to avoid duplication of the logic for the 'isNumber' tests, and the 'isInvalidYear' tests by having some helper functions sort of like this: 

```
export const testInvalidValues = (
	fn: (value: any) => boolean,
	invalidValues: any[],
	expectedResult: boolean,
) => {
	invalidValues.forEach((value) => {
		it(`should return ${expectedResult} for value: ${String(value)}`, () => {
			const result = fn(value);
			expect(result).toEqual(expectedResult);
		});
	});
};

export const testValidValues = (
	fn: (value: any) => boolean,
	validValues: any[],
	expectedResult: boolean,
) => {
	validValues.forEach((value) => {
		it(`should return ${expectedResult} for value: ${String(value)}`, () => {
			const result = fn(value);
			expect(result).toEqual(expectedResult);
		});
	});
};
```

smorrisods:
> This overall looks really great @smorrisods. I'm wondering if there is a way to avoid duplication of the logic for the 'isNumber' tests, and the 'isInvalidYear' tests by having some helper functions sort of like this:

Good catch noticing the unit tests all look very similar.  That is by design.

The goal of the tests is to be atomic and clear in each about what they do.  The duplication is on purpose so that we can easily see what the test is setting up and what it is executing.  Unit tests should test one thing, and one thing only, and they should be clear and easy to understand.

The test pattern I'm following is the _Arrange, Act, and Assert (AAA)_ pattern.  Here is how it works:

_Arrange_
In the arrange step you setup the values you need in order to act out what you're wanting to test.  Things to arrange are, input values, expected values, and any requirements needed up front.

_Act_
In Act is where you perform the action you want to test, like calling a function.  Act should be as short and concise as possible as each unit test should test only one action at a time.

_Assert_
In Assert you validate your results against an expected value or state.

A sample using our testing framework:

```ts
describe('taxCalculator', () => {
	it('should calculate tax amount given a tax percentage', () => {
		// Arrange
		const invoiceSubtotalInDollars = 10;
		const taxPercentage = 13;
		const expectedValueInDollars = 1.3;

		// Act
		const result = calculateTaxAmount(invoiceSubtotalInDollars , taxPercentage);

		// Assert
		expect(result).toEqual(expectedValueInDollars ); // result should be exactly 1.3, as per the expectedValue.
	});
});
```

By following this structure we can see what the input values are, what the execution call is, and what the expected results should be at a glace.  If a test breaks, we don't have to hunt through helper functions to see if there is a bug in them vs. a bug in our code, we can validate and update (of required) only the impacted test cases.

I'm happy to talk more about this.

smorrisods:
> This overall looks really great @smorrisods. I'm wondering if there is a way to avoid duplication of the logic for the 'isNumber' tests, and the 'isInvalidYear' tests by having some helper functions sort of like this:

Good catch noticing the unit tests all look very similar, @erincarrieresontario .  That is by design.

The goal of the tests is to be atomic and clear in each about what they do.  The duplication is on purpose so that we can easily see what the test is setting up and what it is executing.  Unit tests should test one thing, and one thing only, and they should be clear and easy to understand.

The test pattern I'm following is the _Arrange, Act, and Assert (AAA)_ pattern.  Here is how it works:

_Arrange_
In the arrange step you setup the values you need in order to act out what you're wanting to test.  Things to arrange are, input values, expected values, and any requirements needed up front.

_Act_
In Act is where you perform the action you want to test, like calling a function.  Act should be as short and concise as possible as each unit test should test only one action at a time.

_Assert_
In Assert you validate your results against an expected value or state.

A sample using our testing framework:

```ts
describe('taxCalculator', () => {
	it('should calculate tax amount given a tax percentage', () => {
		// Arrange
		const invoiceSubtotalInDollars = 10;
		const taxPercentage = 13;
		const expectedValueInDollars = 1.3;

		// Act
		const result = calculateTaxAmount(invoiceSubtotalInDollars , taxPercentage);

		// Assert
		expect(result).toEqual(expectedValueInDollars ); // result should be exactly 1.3, as per the expectedValue.
	});
});
```

By following this structure we can see what the input values are, what the execution call is, and what the expected results should be at a glace.  If a test breaks, we don't have to hunt through helper functions to see if there is a bug in them vs. a bug in our code, we can validate and update (of required) only the impacted test cases.

I'm happy to talk more about this.

erincarrieresontario:
> > This overall looks really great @smorrisods. I'm wondering if there is a way to avoid duplication of the logic for the 'isNumber' tests, and the 'isInvalidYear' tests by having some helper functions sort of like this:
> 
> Good catch noticing the unit tests all look very similar, @erincarrieresontario . That is by design.
> 
> The goal of the tests is to be atomic and clear in each about what they do. The duplication is on purpose so that we can easily see what the test is setting up and what it is executing. Unit tests should test one thing, and one thing only, and they should be clear and easy to understand.
> 
> The test pattern I'm following is the _Arrange, Act, and Assert (AAA)_ pattern. Here is how it works:
> 
> _Arrange_ In the arrange step you setup the values you need in order to act out what you're wanting to test. Things to arrange are, input values, expected values, and any requirements needed up front.
> 
> _Act_ In Act is where you perform the action you want to test, like calling a function. Act should be as short and concise as possible as each unit test should test only one action at a time.
> 
> _Assert_ In Assert you validate your results against an expected value or state.
> 
> A sample using our testing framework:
> 
> ```ts
> describe('taxCalculator', () => {
> 	it('should calculate tax amount given a tax percentage', () => {
> 		// Arrange
> 		const invoiceSubtotalInDollars = 10;
> 		const taxPercentage = 13;
> 		const expectedValueInDollars = 1.3;
> 
> 		// Act
> 		const result = calculateTaxAmount(invoiceSubtotalInDollars , taxPercentage);
> 
> 		// Assert
> 		expect(result).toEqual(expectedValueInDollars ); // result should be exactly 1.3, as per the expectedValue.
> 	});
> });
> ```
> 
> By following this structure we can see what the input values are, what the execution call is, and what the expected results should be at a glace. If a test breaks, we don't have to hunt through helper functions to see if there is a bug in them vs. a bug in our code, we can validate and update (of required) only the impacted test cases.
> 
> I'm happy to talk more about this.

Cool - that does make sense! Thanks for explaining :)

erincarrieresontario (approved):

PR #9, opened 2024-12-16
DS-1933: Removes margins from form web components

MeaghanCarrieres:
- Removes margins from "ontario-input", "ontario-checkboxes", "ontario-button" and "ontario-textarea" web components
- Removes margins from ontario-input class in text-input component styles in global styles package to target "ontario-date-input" margins

### DIFF
--- a/packages/ontario-design-system-component-library/src/components/ontario-button/ontario-button.scss
+++ b/packages/ontario-design-system-component-library/src/components/ontario-button/ontario-button.scss
@@ -25,7 +25,7 @@ $ontario-button-bg-secondary: colours.$ontario-colour-white;
 	font-family: typography.$ontario-font-open-sans;
 	font-weight: fontWeights.$ontario-font-weights-semi-bold;
 	line-height: math.div(14, 9);
-	margin: spacing.$spacing-0 (spacing.$spacing-4 + spacing.$spacing-3) spacing.$spacing-5 spacing.$spacing-0;
+	margin: spacing.$spacing-0;
 	min-width: 10rem;
 	padding: math.div((spacing.$spacing-4 + spacing.$spacing-1), 2) spacing.$spacing-5;
 	text-align: center;
--- a/packages/ontario-design-system-component-library/src/components/ontario-checkbox/ontario-checkboxes.scss
+++ b/packages/ontario-design-system-component-library/src/components/ontario-checkbox/ontario-checkboxes.scss
@@ -22,7 +22,6 @@ $ontario-checkbox-box-shadow-outline: globalFunctions.px-to-rem(4);
 }
 
 .ontario-checkboxes {
-	margin-bottom: spacing.$spacing-7;
 	max-width: globalVariables.$standard-width;
 }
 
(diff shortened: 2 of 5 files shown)


matt-stjean:
@MeaghanCarrieres Does anything need to be done for `ontario-dropdown-list` and `ontario-radio-buttons`? I see them in the ticket but not the changes.

MeaghanCarrieres:
> @MeaghanCarrieres Does anything need to be done for `ontario-dropdown-list` and `ontario-radio-buttons`? I see them in the ticket but not the changes.

Hey @matt-stjean! The `ontario-dropdown-list` component makes use of the `ontario-input` class, so when the margins were removed from that, it fixed it.

As for the radio buttons, they don't seem to have any margins on their group, and only on individual radio button items themselves. The margin that is causing the space at the bottom of their grouping is from the `ontario-form-group` class, which I believe will be removed/refactored in another ticket in this epic.

matt-stjean (commented):
Thanks for the reply @MeaghanCarrieres, it looks good to me.

smorrisods:
Looks like this is what is adding the extra spacing.  I think this rule will not be needed with the overall changes to how we handle form spacing as the vertical spacing.  I think we should clean this up in this PR as it is relevant to the thematic changes being made.

From: `packages/ontario-design-system-global-styles/src/styles/scss/6-components/_text-inputs.component.scss`

```sass 
.ontario-form-group:last-of-type {
	margin-bottom: 2.5rem;
}
```

MeaghanCarrieres:
> Looks like this is what is adding the extra spacing. I think this rule will not be needed with the overall changes to how we handle form spacing as the vertical spacing. I think we should clean this up in this PR as it is relevant to the thematic changes being made.
> 
> From: `packages/ontario-design-system-global-styles/src/styles/scss/6-components/_text-inputs.component.scss`
> 
> ```css-scss
> .ontario-form-group:last-of-type {
> 	margin-bottom: 2.5rem;
> }
> ```
> 
> It seems to also show up in `packages/ontario-design-system-global-styles/src/styles/scss/6-components/_form.component.scss`:
> 
> ```css-scss
> .ontario-form-group:last-of-type {
> 	margin-bottom: 2.5rem;
> }
> 
> .ontario-form-group:last-of-type {
> 	margin-bottom: spacing.$spacing-8;
> }
> ```
> 
> We should figure out what to do with the `ontario-form-group` class, as it impacts the distribution package. Any thoughts?

Hey Scott,
I removed the styles for now, but now that I'm investigating its use, I'm not sure if we need it at all. I originally thought it was used just to contain the labels and inputs, but it seems like those containers are already there for other classes (such as ones to indicate errors or hint expanders, etc.). So, it seems like we can just get rid of it alltogether?

smorrisods:
I wonder how this `ontario-form-group` SASS translates into the dist package 🤔

MeaghanCarrieres:
> > So, it seems like we can just get rid of it alltogether?
> 
> This was the sense I got too a while back, but it needs more detailed exploration to make sure we're not causing carnage by removing it. 🐱

I agree, it should be looked into more. Does it make sense to move that work into a new ticket within the epic, and I'll revert the styles back for this branch?

smorrisods:
> I agree, it should be looked into more. Does it make sense to move that work into a new ticket within the epic, and I'll revert the ontario-form-group styles back for this branch?

That seems reasonable given the weird interconnected nature of these files.  Let's do that.

smorrisods:
Looks good to me!

PR #10, opened 2024-12-16
DS-2067: declare $modules before use

smorrisods:
This avoids build errors in some bundlers, eg. with NextJS.

These changes have been tested by Dragan in his application and they compiled successfully.

Closes DS-2067


kyle-binger-3:
Looks good to me @smorrisods!

PR #16, opened 2025-01-09
DS-2113: Move component styles within global styles project to component library

MeaghanCarrieres:
This work refactors component styles in both the `global-styles` and `component-library` packages.

- Moves component specific styles from `global-styles` package to `component-library` package
- Updates affected web components `.scss` files to reference correct new paths for component styles.
- Adds output target in Stencil config of `component-library` package to generate theme SCSS file in the component library distribution package.
- Adds slotted styles folder in `component-library` package. This folder as well as the `global.scss` file are copied in both angular + react component libraries through their respective `package.json` files.
- Guidance on how to include support for slotted styles was added in the react component library + angular component library readme's
- Updates to `app-react` and `app-angular` project to include the new theme into the files for slotted styles to work.

Closes DS-2113


smorrisods (changes requested):
Only one little change I noticed.

erincarrieresontario (approved):
Looks great!

packages/ontario-design-system-component-library/src/components/ontario-footer/ontario-footer.scss
```diff
@@ -5,8 +5,9 @@
 @use '@ongov/ontario-design-system-global-styles/dist/styles/scss/1-variables/breakpoints.variables' as breakpoints;
 @use '@ongov/ontario-design-system-global-styles/dist/styles/scss/1-variables/typography.variables' as typography;
 @use '@ongov/ontario-design-system-global-styles/dist/styles/scss/1-variables/font-weights.variables' as fontWeights;
-@use '../../styles/grid.scss';
-@use '../../styles/common.scss';
+@use '@ongov/ontario-design-system-global-styles/dist/styles/scss/4-elements/_global.elements.scss';
+@use '@ongov/ontario-design-system-global-styles/dist/styles/scss/5-layout/_grid.layout.scss';
+@forward '@ongov/ontario-design-system-global-styles/dist/styles/scss/7-overrides/spacing.overrides.scss';
```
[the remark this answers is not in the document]
MeaghanCarrieres: Oh! yes it should. good catch!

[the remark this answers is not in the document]
MeaghanCarrieres: actually, i think it is using forward to not duplicate the spacing variable

packages/ontario-design-system-component-library-react/package.json
```diff
@@ -20,10 +20,12 @@
 		"clean": "rm -rf dist",
 		"clean:full": "rm -rf node_modules && npm run clean",
 		"compile": "npm run tsc",
-		"copy:assets": "npm run copy:images && npm run copy:favicons && npm run copy:fonts",
+		"copy:assets": "npm run copy:images && npm run copy:favicons && npm run copy:fonts && npm run copy:theme && npm run copy:slotted-styles",
 		"copy:favicons": "copyfiles -E -f \"node_modules/@ongov/ontario-design-system-component-library/dist/ontario-design-system-components/favicons/**/*\" dist/assets/favicons",
 		"copy:fonts": "copyfiles -E -u 6 \"node_modules/@ongov/ontario-design-system-component-library/dist/ontario-design-system-components/fonts/**/*\" dist/assets/fonts",
 		"copy:images": "copyfiles -E -f \"node_modules/@ongov/ontario-design-system-component-library/dist/ontario-design-system-components/assets/**/*\" dist/assets/images",
+		"copy:slotted-styles": "copyfiles -E -f \"node_modules/@ongov/ontario-design-system-component-library/dist/ontario-design-system-components/styles/**/*\" dist/styles/slotted-styles",
+		"copy:theme": "copyfiles -E -f \"node_modules/@ongov/ontario-design-system-component-library/dist/ontario-design-system-components/theme.scss\" dist",
```
smorrisods, line 28: This looks to copy the `theme.scss` file to the main `dist` folder instead of the `styles` folder, is there a purpose for this location?

image

packages/ontario-design-system-component-library-react/package.json
smorrisods, line 1: Around the 

```json
	"main": "./dist/index.js",
	"module": "./dist/index.js",
	"types": "./dist/index.d.ts",
```

section we should add  the `theme.scss`, if this is the correct file, to a `"style"` section here in the `package.json` file.  The same goes for the Angular App.

packages/ontario-design-system-component-library/src/styles/slotted-styles/page-alerts.scss
```diff
@@ -0,0 +1,43 @@
...
+		color: colours.$ontario-colour-alert;
+
+		&:hover,
+		&:focus {
+			color: #6b0000;
+		}
+
+		&:active {
+			color: #280000;
+		}
```
smorrisods, line 41: We should name these colours as SASS variables in this file so they have some meaning behind them.

packages/ontario-design-system-component-library/stencil.config.ts
```diff
@@ -52,6 +52,14 @@ export const config: Config = {
 					src: 'translations/*.i18n.json',
 					dest: 'i18n',
 				},
+				{
+					src: '../src/global.scss',
+					dest: 'theme.scss',
```
smorrisods, line 57: I think this should also be in the `styles` folder.

packages/ontario-design-system-global-styles/src/styles/scss/theme.scss
```diff
@@ -47,13 +47,8 @@
...
 @forward './6-components/buttons.component';
-@forward './6-components/form.component';
-@forward './6-components/labels.component';
-@forward './6-components/text-inputs.component';
-@forward './6-components/callouts-asides';
-@forward './6-components/page-alerts.component';
 @forward './6-components/all.component';
 
-/*** 7 - Overrides ***/
+/*** 7 - Overrides *å**/
```
smorrisods, line 52: A special character seems to have been introduced here.

```suggestion
/*** 7 - Overrides ***/
```

smorrisods (changes requested):
Looking good, I added some minor question/comments.

packages/ontario-design-system-component-library-angular/README.md
```diff
@@ -23,7 +23,25 @@ To use the Ontario Design System Angular component library, follow these steps:
...
+     "options": {
+        "styles": [node_modules/@ongov/ontario-design-system-component-library-angular/dist/theme.scss]
+     }
+  }
+  ```
+
+- Forward the theme in your main stylesheet:
+
+  ```scss
+  @use '@ongov/ontario-design-system-component-library-angular/dist/theme.scss';
```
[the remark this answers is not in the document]
MeaghanCarrieres: Good catch! updated.

packages/ontario-design-system-component-library/src/styles/theme.scss
smorrisods, line 1: Since this is autogenerated, we should not include it in the repositoy and have it git ignored to avoid creating dirty copies when running the build. These will block releasing if git decides something has changed that doesn't match what is in the repo.

packages/ontario-design-system-component-library/stencil.config.ts
```diff
@@ -52,6 +52,14 @@ export const config: Config = {
...
 					dest: 'i18n',
 				},
+				{
+					src: '../src/styles/theme.scss',
+					dest: 'styles/theme.scss',
+				},
+				{
+					src: '../src/styles/slotted-styles',
+					dest: 'styles/slotted-styles',
+				},
```
smorrisods, line 62: It looks the `copy` feature can handle copying folders, perhaps this can be simplified by copying over the `styles` folder?

smorrisods (changes requested):
I have left a few minor ideas, but it's coming together nicely 😁

packages/ontario-design-system-component-library/stencil.config.ts
MeaghanCarrieres: Hey Scott! I actually specifically only wanted to copy over the slotted styles folder because the other files in this folder are for the components, which are bundled with the component's SASS files.

packages/ontario-design-system-component-library/scripts/build-theme-scss.ts
```diff
@@ -0,0 +1,37 @@
...
+	`// Changes will be overwritten during the build process.\n\n` +
+	`@forward '@ongov/ontario-design-system-global-styles/dist/styles/scss/theme.scss';\n\n`;
+
+// Get all .scss files from the source folder
+const scssFiles: string[] = fs.readdirSync(sourceFolder).filter((file: string) => file.endsWith('.scss'));
+
+if (scssFiles.length === 0) {
+	console.error('No SCSS files found in the source folder.');
+	process.exit(1);
+}
```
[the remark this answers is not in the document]
MeaghanCarrieres: I think my thought process here was that it could help risk generating an incomplete theme (like missing some imports for example) without noticing, but maybe its a bit overkill? I guess we could update it to have a `console.warn` instead. What do you think?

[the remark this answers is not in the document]
smorrisods: I think we should continue if there's nothing there - it is just aggregating whatever we put in those folders, so if we redo this in the future this could cause it to fail.  If the assumption is that there is always supposed to be something there then I like your idea of a warning, otherwise, we can probably just go with a simple console log.

Validation of the theme file maybe should be a integration test 🤔.  What do you think?

packages/ontario-design-system-component-library/scripts/build-theme-scss.ts
```diff
@@ -0,0 +1,39 @@
...
+
+// Add imports for each file
+scssFiles.forEach((file: string) => {
+	const relativePath: string = `./${path
+		.relative(outputFolder, path.join(slottedStyleSourceFolder, file))
+		.replace(/\\/g, '/')}`;
+	scssContent += `@use '${relativePath}';\n`;
```
smorrisods, line 33: Should these additional styles be included as unnamed `@use` imports or just forwarded through with `@forward`?

MeaghanCarrieres: The imports should be using `@use`!

packages/ontario-design-system-component-library/scripts/build-theme-scss.ts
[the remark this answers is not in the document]
MeaghanCarrieres: Tests sound good! I've created a new issue to address writing it: https://ontariodigital.atlassian.net/browse/DS-2147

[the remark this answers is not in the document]
MeaghanCarrieres: I've also updated this to be a warn rather than error + exit.

smorrisods (approved):
This looks good!

PR #21, opened 2025-01-17
DS-2111: cards not full width

erincarrieresontario:
Fix for DS-2111: Update widths of ontario-card-collection rows to account for gap between cards. This allows the cards to span the full width of the ontario-row.

### DIFF
--- a/packages/ontario-design-system-component-library/src/components/ontario-card-collection/ontario-card-collection.scss
+++ b/packages/ontario-design-system-component-library/src/components/ontario-card-collection/ontario-card-collection.scss
@@ -23,7 +23,7 @@
 }
 
 .ontario-card-collecton--cards-per-row-3 ::slotted(ontario-card) {
-	width: calc(33.3% - 1.25rem);
+	width: calc(33.3% - 1.33rem);
 
 	@media screen and (max-width: breakpoints.$small-breakpoint) {
 		width: 100;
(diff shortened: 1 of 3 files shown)


packages/ontario-design-system-component-library/src/components/ontario-card-collection/ontario-card-collection.scss
```diff
@@ -5,51 +5,35 @@
...
-	width: calc(33.3% - 1.75rem);
-
-	&:nth-child(3n) {
-		margin-right: spacing.$spacing-0;
-	}
+	width: calc(33.3% - 1.33rem);
 
 	@media screen and (max-width: breakpoints.$small-breakpoint) {
-		width: calc(100% - 0.5rem);
+		width: 100;
```
smorrisods, line 29: Is this supposed to be `100%`?

```suggestion
		width: 100%;
```

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.scss
```diff
@@ -186,7 +190,7 @@ $ontario-card-heading-colours: (
 
 .ontario-card__text-container {
 	.ontario-card__card-type--horizontal & {
-		width: 66.6%;
+		width: 66.66%;
```
smorrisods, line 193: Seeing as this is a repeating decimal, using it's fractional form might yield a more consistent result.

```suggestion
		width: math.percentage(2/3);
```

smorrisods (changes requested):
Looks pretty good 🙂

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.scss
smorrisods: Good catch on the use of `math.div`, SASS has been busy changing things haha

smorrisods (approved):

PR #27, opened 2025-02-10
fix: fix broken symlinks in packages

MeaghanCarrieres:
This branch fixes broken symlinks in project by setting the `link-workspace-packages` to true in the `.npmrc` file. 

It also makes sure the react component library is using the exact version of the component library package in its dependencies.


smorrisods (approved):
LGTM

PR #28, opened 2025-02-19
DS-2105: Creates ontario-form-container web component

MeaghanCarrieres:
This MR introduces a new component for vertical form spacing and improves spacing management within form elements.

Changes include:

- Added `ontario-form-container` component with e2e and unit tests to the **component library**. (Note that the README for this component will be updated in [DS-1937](https://ontariodigital.atlassian.net/browse/DS-1937))
- Removed vertical form spacing styles from the **global-styles package** and moved them into the `ontario-form-container` styles.
- Updated the `ontario-fieldset` component to detect if it's nested within an `ontario-form-container` component and apply the appropriate spacing value via slotted styles.

[DS-1937]: https://ontariodigital.atlassian.net/browse/DS-1937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/ontario-design-system-component-library/src/components/ontario-fieldset/ontario-fieldset.scss
```diff
@@ -1,6 +1,33 @@
+@use '@ongov/ontario-design-system-global-styles/dist/styles/scss/1-variables/spacing.variables' as spacing;
 @use '@ongov/ontario-design-system-global-styles/dist/styles/scss/4-elements/_global.elements.scss';
 
 @use '../../styles/fieldsets.scss';
 @use '../../styles/labels.scss';
 @use '../../styles/text-inputs.scss';
 @use '@ongov/ontario-design-system-global-styles/dist/styles/scss/7-overrides/_visibility.overrides.scss';
+
+.ontario-fieldset--spacing-40 {
```
matt-stjean, line 9: I am very unfamiliar with the work for this PR, so forgive some of my questions. But why did we choose "-40" and "-16" as our two offered options?

packages/ontario-design-system-component-library/src/components/ontario-fieldset/ontario-fieldset.scss
```diff
@@ -1,6 +1,33 @@
...
+	::slotted(ontario-input:not(:last-child)),
+	::slotted(ontario-checkboxes:not(:last-child)),
+	::slotted(ontario-dropdown-list:not(:last-child)),
+	::slotted(ontario-textarea:not(:last-child)),
+	::slotted(ontario-radio-buttons:not(:last-child)),
+	::slotted(ontario-button:not(:last-child)),
+	::slotted(ontario-date-input:not(:last-child)) {
+		margin-bottom: spacing.$spacing-7;
```
matt-stjean, line 17: Similar to the above question, just wondering why specifically we chose to use the "spacing-7" and "spacing-4" for the margins here. Do we have a requirement for how much spacing there was supposed to be?

packages/ontario-design-system-component-library/src/components/ontario-fieldset/ontario-fieldset.scss
```diff
@@ -1,6 +1,33 @@
...
+	::slotted(ontario-input:not(:last-child)),
+	::slotted(ontario-checkboxes:not(:last-child)),
+	::slotted(ontario-dropdown-list:not(:last-child)),
+	::slotted(ontario-textarea:not(:last-child)),
+	::slotted(ontario-radio-buttons:not(:last-child)),
+	::slotted(ontario-button:not(:last-child)),
+	::slotted(ontario-date-input:not(:last-child)) {
+		margin-bottom: spacing.$spacing-4;
+		display: block;
+	}
```
matt-stjean, line 32: When we have a new form element, they'll have to be added here right? How can we mitigate the chance we miss updating this spot?

packages/ontario-design-system-component-library/src/components/ontario-fieldset/ontario-fieldset.tsx
```diff
@@ -22,6 +24,13 @@ export class OntarioFieldset implements Fieldset {
 	 */
 	@Prop({ mutable: true }) legendSize: CaptionType = 'default';
 
+	/**
+	 * The following states are used to determine if the parent is a `ontario-form-container` component.
+	 * This is to help style the form children if they are contained within a fieldset.
+	 */
+	@State() hasFormContainerParent: boolean = false;
+	@State() gapValue: number = 40;
```
matt-stjean, line 32: the use of "40" for gapValue feels a bit like a magic number without context. It might be worth adding to the comment how this helps in the case form children are in a fieldset. Maybe overkill but it might help understanding implication

packages/ontario-design-system-component-library/src/components/ontario-fieldset/ontario-fieldset.tsx
```diff
@@ -69,12 +78,20 @@ export class OntarioFieldset implements Fieldset {
...
+
 	componentWillLoad() {
 		this.validateLegend();
 		this.validateLegendSize();
+
+		// Check if the direct or any ancestor parent is `ontario-form-container`
+		this.hasFormContainerParent = this.element.closest('ontario-form-container') !== null;
+		const formContainer = this.element.closest('ontario-form-container') as HTMLElement | null;
+		if (formContainer) {
+			this.gapValue = (formContainer as any).gap === 'condensed' ? 16 : 40;
```
matt-stjean, line 90: Similar comment to before, 16 and 40 seem arbitrary to me without context. Can we swap these hardcoded values with a variable? I am seeing 16 and 40 a couple times in this page.

packages/ontario-design-system-component-library/src/components/ontario-fieldset/ontario-fieldset.tsx
```diff
@@ -86,10 +103,20 @@ export class OntarioFieldset implements Fieldset {
 		}
 	}
 
+	private getFieldsetClass() {
+		let baseClass = 'ontario-fieldset';
+		if (this.hasFormContainerParent && this.gapValue) {
```
matt-stjean, line 108: would it be worth checking that gapValue === 16 or 40 here? Instead of just that it exists? Or are we confident that if it exists, it will have the value we need?

packages/ontario-design-system-component-library/src/components/ontario-fieldset/ontario-fieldset.tsx
```diff
@@ -86,10 +103,20 @@ export class OntarioFieldset implements Fieldset {
...
+			baseClass += `ontario-fieldset ontario-fieldset--gap-${this.gapValue}`;
```
matt-stjean, line 109: if baseClass = 'ontario-fieldset', then it looks like this line will add a second instance of 'ontario-fieldset' resulting in a class list like:
`ontario-fieldset ontario-fieldset ontario-fieldset--gap-XX'

packages/ontario-design-system-component-library/src/components/ontario-form-container/ontario-form-container.scss
```diff
@@ -0,0 +1,77 @@
+@use '@ongov/ontario-design-system-global-styles/dist/styles/scss/1-variables/spacing.variables' as spacing;
+
+// Mixin for generating individual form container gap classes
+@mixin ontario-form-container-gap($modifier, $spacing-variable) {
+	.ontario-form-container--gap-#{$modifier},
+	.ontario-form-container--gap-#{$modifier} form {
+		::slotted(ontario-input:not(:last-child)),
```
matt-stjean, line 7: Similar worry to the ontario-fieldset.scss, we will need to update this if any new form elements are created.

packages/ontario-design-system-component-library/src/components/ontario-form-container/ontario-form-container.scss
```diff
@@ -0,0 +1,77 @@
...
+					margin-bottom: $child-spacing-variable + $parent-spacing;
+					display: block;
+				}
+			}
+		}
+	}
+}
+
+// Gap definitions
+$gap-definitions: (16, spacing.$spacing-4), (40, spacing.$spacing-7);
```
matt-stjean, line 58: Could use an explanation of the why these values are chosen

matt-stjean (dismissed):
Looks good, Meaghan. Added some comments throughout but pretty minor - mostly looking at future proofing and readability

packages/ontario-design-system-component-library/src/components/ontario-fieldset/ontario-fieldset.scss
MeaghanCarrieres: Hey @matt-stjean, good question - so for now the design recommendation is to only offer either 16px or 40px of bottom margin, which corresponds to the `default` or `condensed` spacing.

Perhaps adding a comment here would help clarify that?

packages/ontario-design-system-component-library/src/components/ontario-fieldset/ontario-fieldset.scss
MeaghanCarrieres: I think my comment above also answers this question!

packages/ontario-design-system-component-library/src/components/ontario-fieldset/ontario-fieldset.scss
MeaghanCarrieres: That's a good point. I think the idea behind this is that only form related components would be affected by this new component, but right now, yes, if a new form element were released you'd have to remember to move it here. I'm wondering how/if we could streamline this a bit.

Maybe if there was a file where the components were stored in a sass variable, and using `@each` to loop over them? 🤔

packages/ontario-design-system-component-library/src/components/ontario-fieldset/ontario-fieldset.tsx
MeaghanCarrieres: Good catch!

packages/ontario-design-system-component-library/src/components/ontario-fieldset/ontario-fieldset.tsx
MeaghanCarrieres: oh damn! good catch!

packages/ontario-design-system-component-library/src/components/ontario-fieldset/ontario-fieldset.tsx
smorrisods: I thought I added a comment about using a named enum with a numerical value, but maybe it got lost in a review I forget to submit.

I recommend creating a single source of truth for these spacing values to avoid using magic numbers all over the place.  This could an named enum with numerical values.  You could probably go as far as typing this state that of the enum rather than just a generic `number` to keep the strong typing going.

packages/ontario-design-system-component-library/src/components/ontario-fieldset/ontario-fieldset.tsx
```diff
@@ -69,12 +87,20 @@ export class OntarioFieldset implements Fieldset {
...
+
 	componentWillLoad() {
 		this.validateLegend();
 		this.validateLegendSize();
+
+		// Check if the direct or any ancestor parent is `ontario-form-container`
+		this.hasFormContainerParent = this.element.closest('ontario-form-container') !== null;
+		const formContainer = this.element.closest('ontario-form-container') as HTMLElement | null;
+		if (formContainer) {
+			this.gapValue = (formContainer as any).gap === 'condensed' ? FormGapCondensed : FormGapDefault;
```
smorrisods, line 99: Same here, we should use an enum for these definitions/type so we don't rely on magic strings.  If you don't want to use an enum then use a type with these two options as the type, `default | condensed`.

As for this explicit any, we know the type of `formContainer` so we should be able to say `formContainer as OntarioFormContainer` and keep the strong typing for the `gap` property.  This will help keep the code coherent and reactors should be easier because we're tracking where we're using these properties.

smorrisods: This enum could even use the values (like the comment on `gapValue` so we don't have to define them as separeate entities.  This might make using an enum a better choice than just a `type`.

packages/ontario-design-system-component-library/src/components/ontario-form-container/ontario-form-container.tsx
```diff
@@ -0,0 +1,38 @@
...
+	/**
+	 * Defines the gap (bottom margin) between slotted form elements.
+	 * If no gap prop is provided, it will default to 'default'.
+	 */
+	@Prop() gap: 'default' | 'condensed' = 'default';
+
+	/**
+	 * Returns the pixel value for the selected gap option.
+	 */
+	private getGapValue() {
```
smorrisods, line 22: This function has no explicit return type to restrict the values

packages/ontario-design-system-component-library/src/components/ontario-form-container/readme.md
```diff
@@ -0,0 +1,13 @@
+# ontario-form-container
+
```
smorrisods, line 2: We'll want documentation here prior to merging so folks know how to use it.

packages/ontario-design-system-component-library/src/components/ontario-form-container/test/__snapshots__/ontario-form-container.spec.tsx.snap
```diff
@@ -0,0 +1,12 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`ontario-form-container snapshot should render the expected html 1`] = `
+<ontario-form-container>
+  <template shadowrootmode="open">
+    <div class="ontario-form-container--gap-40">
+      <slot></slot>
+    </div>
+  </template>
+  <slot></slot>
```
smorrisods, line 10: I wonder what this additional `<slot>` is from?

packages/ontario-design-system-component-library/src/components/ontario-form-container/test/ontario-form-container.spec.tsx
```diff
@@ -0,0 +1,32 @@
+import { newSpecPage } from '@stencil/core/testing';
+import { OntarioFormContainer } from '../ontario-form-container';
+
+describe('ontario-form-container', () => {
+	describe('snapshot', () => {
+		it('should render the expected html', async () => {
+			const page = await newSpecPage({
+				components: [OntarioFormContainer],
+				html: `<ontario-form-container><slot></slot></ontario-form-container>`,
```
smorrisods, line 9: Oh, this is where the additional `<slot>` is from? Is it normal to use a `<slot>` in regular HTML that isn't a component?

smorrisods (changes requested):
I've added a few more notes, mostly to try and tighten up where data is from and how it is connected across the components.

packages/ontario-design-system-component-library/src/components/ontario-form-container/readme.md
MeaghanCarrieres: Hey @smorrisods, the documentation will be part of [DS-1937](https://ontariodigital.atlassian.net/browse/DS-1937).

packages/ontario-design-system-component-library/src/components/ontario-form-container/test/ontario-form-container.spec.tsx
MeaghanCarrieres: Good catch! I don't think the `<slot>` should be included in this html line.

packages/ontario-design-system-component-library/src/components/ontario-fieldset/ontario-fieldset.scss
```diff
@@ -1,6 +1,41 @@
...
 @use '@ongov/ontario-design-system-global-styles/dist/styles/scss/4-elements/_global.elements.scss';
 
 @use '../../styles/fieldsets.scss';
 @use '../../styles/labels.scss';
 @use '../../styles/text-inputs.scss';
+
 @use '@ongov/ontario-design-system-global-styles/dist/styles/scss/7-overrides/_visibility.overrides.scss';
+
+/**
+ * The following classes define the gap (bottom margin) between slotted form elements.
```
erincarrieresontario, line 11: Should we state that this is in px?

erincarrieresontario (approved):
This looks good Meaghan!

packages/ontario-design-system-component-library/src/components/ontario-fieldset/ontario-fieldset.scss
MeaghanCarrieres: Updated!

packages/ontario-design-system-component-library/src/components/ontario-fieldset/ontario-fieldset.tsx
```diff
@@ -69,12 +84,20 @@ export class OntarioFieldset implements Fieldset {
...
 	componentWillLoad() {
 		this.validateLegend();
 		this.validateLegendSize();
+
+		// Check if the direct or any ancestor parent is `ontario-form-container`
+		this.hasFormContainerParent = this.element.closest('ontario-form-container') !== null;
+		const formContainer = this.element.closest('ontario-form-container') as OntarioFormContainer | null;
+		if (formContainer) {
+			this.gapValue = formContainer.gap === 'condensed' ? FormGap.Condensed : FormGap.Default;
+		}
```
smorrisods, line 97: I recommend moving this logic to a function

smorrisods (changes requested):

packages/ontario-design-system-component-library/src/components/ontario-fieldset/ontario-fieldset.scss
matt-stjean: That is a great solution

matt-stjean:
Looks great @MeaghanCarrieres

matt-stjean (approved):

PR #31, opened 2025-03-04
DS-2148: header overflow menu unit tests

erincarrieresontario:
- This MR includes fixes and improvements to the ontario-header component and its corresponding tests, along with new tests for the ontario-header-overflow-menu component.
- Updated ontario-header spec test to account for the new 'ontario-header-overflow-menu' component
- Added spec and e2e tests for the 'ontario-header-overflow-menu' component


packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/test/ontario-header-overflow-menu.spec.tsx
```diff
@@ -0,0 +1,75 @@
...
+	it('should toggle menu visibility when the menuButtonToggled event is received', async () => {
+		const menuItems = JSON.stringify([{ title: 'Link 1', href: '/link-1', linkIsActive: false }]);
+
+		const page = await newSpecPage({
+			components: [OntarioHeaderApplicationMenu],
+			html: `<ontario-header-overflow-menu menu-items='${menuItems}'></ontario-header-overflow-menu>`,
+		});
+
+		expect(page.rootInstance).toBeTruthy();
+		const component = page.rootInstance!;
```
MeaghanCarrieres, line 44: This and the next test seem to use `page.rootInstance` whereas the two above are just using `page.root` - is there a reason they are different?

I also wonder if this is needed if the first test checks for this already.

packages/ontario-design-system-component-library/src/components/ontario-header/readme.md
```diff
@@ -305,14 +305,12 @@ To ensure best practices, it is important to limit the number of navigation link
 ### Depends on
 
 - [ontario-language-toggle](../ontario-language-toggle)
-- [ontario-header-menu](../ontario-header-menu)
```
smorrisods, line 308: Wait, why is it removing this component from the documentation?

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/test/ontario-header-overflow-menu.spec.tsx
```diff
@@ -0,0 +1,75 @@
...
+			{ title: 'Link 1', href: '/link-1', linkIsActive: false },
+			{ title: 'Link 2', href: '/link-2', linkIsActive: false },
+		]);
+
+		const page = await newSpecPage({
+			components: [OntarioHeaderApplicationMenu],
+			html: `<ontario-header-overflow-menu menu-items='${menuItems}'></ontario-header-overflow-menu>`,
+		});
+
+		expect(page.root).toBeTruthy();
```
smorrisods, line 25: This is currently under test in `should render without errors` so we probably don't need it repeated.

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/test/ontario-header-overflow-menu.spec.tsx
```diff
@@ -0,0 +1,75 @@
...
+		const menuLinks = page.root?.shadowRoot?.querySelectorAll('a');
+
+		expect(menuLinks?.length).toBe(2);
+		expect(menuLinks?.[0]?.textContent).toBe('Link 1');
+		expect(menuLinks?.[0]?.getAttribute('href')).toBe('/link-1');
+		expect(menuLinks?.[1]?.textContent).toBe('Link 2');
+		expect(menuLinks?.[1]?.getAttribute('href')).toBe('/link-2');
```
smorrisods, line 32: I think this is a good candidate for a snapshot, if the links aren't properly rendered the HTML will be broken.

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/test/ontario-header-overflow-menu.spec.tsx
```diff
@@ -0,0 +1,75 @@
...
+	});
+
+	it('should toggle menu visibility when the menuButtonToggled event is received', async () => {
```
smorrisods, line 35: This can be broken into two tests, one toggling the menu open with the event and one toggling it closed with the event.

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/test/ontario-header-overflow-menu.spec.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+	it('should close menu when clicking outside', async () => {
+		const menuItems = JSON.stringify([{ title: 'Link 1', href: '/link-1', linkIsActive: false }]);
+
+		const page = await newSpecPage({
+			components: [OntarioHeaderApplicationMenu],
+			html: `<ontario-header-overflow-menu menu-items='${menuItems}'></ontario-header-overflow-menu>`,
+		});
+
+		expect(page.rootInstance).toBeTruthy();
```
smorrisods, line 63: This is tested a number of times, so it can probably be omitted and let the test fail naturally if it isn't rendered.

smorrisods (changes requested):
Thanks for adding tests! I've added a few comments about breaking down the cases a bit further. I wonder if we should test the focus state in a few ways, including if it moves with hitting the tab key?

erincarrieresontario:
> Thanks for adding tests! I've added a few comments about breaking down the cases a bit further. I wonder if we should test the focus state in a few ways, including if it moves with hitting the tab key?

You're probably right. I just added some accessibility work for the focus states, so that will be merged in with a new branch. Once that is merged in I could probably add some testing for that.

smorrisods (approved):
LGTM

PR #33, opened 2025-03-07
DS-2141: updated and reorganized header styles to remove redundancy

erincarrieresontario:
- Re-organized ontario-header.scss, ontario-header-overflow-menu.scss, ontario-application-header.scss, and service-ontario-header.scss
- Added common header styles to the global styles package.


packages/ontario-design-system-global-styles/src/styles/scss/6-components/_header.component.scss
smorrisods, line 1: I'm going to put a warning here that we should avoid moving the component styles too far away from the components themselves.

These changes should live in the component library as they are not part of the core theme of the global styles.  Recently the form styles were moved out of the global styles into the `src/styles` folder within the component library and they're packaged into the complete styles package by the build process.  Checkout how they're done and reach out to @MeaghanCarrieres for more details in this shift in architecture.

Layer 6 is now generally available within the component library.

smorrisods (changes requested):

packages/ontario-design-system-global-styles/src/styles/scss/6-components/_header.component.scss
MeaghanCarrieres: Hey @erincarrieresontario, some of the work I did for the `ontario-form-container` involved moving out some styles that were more component specific from the global styles package into the component-library package. You can see this work in my [DS-1932 branch](https://github.com/ongov/ontario-design-system/tree/DS-1932/vertical-form-element-spacing/packages/ontario-design-system-component-library/src/styles). I think to continue with that new structure, you could follow this! LMK if you have questions

MeaghanCarrieres:
Hey @erincarrieresontario (and cc: @kyle-binger-3), these styles changes look good to me, but I am noticing a few bugs when I run the component library index file on your branch. Granted, I don't know if these are a result of these changes or perhaps something in the base branch, but think they should be pointed out regardless. I don't see these issues when i run the `develop` branch.

1. Both the ServiceOntario & Ontario header styles seem to be broken:

Service Ontario:
Screenshot 2025-03-18 at 10 24 56 AM

Ontario:
Screenshot 2025-03-18 at 11 08 37 AM

2. The application menu title + menu button don't seem to have the correct focus/active styles:

Screenshot 2025-03-18 at 10 02 10 AM
Screenshot 2025-03-18 at 10 02 17 AM

3. I think this is unrelated to your changes here, but I notice the application menu seems to be displaying the default dynamic menu items when the `maxSubheaderLinks` are set, but no `menuLinks` are set. I'm not sure if it is meant to be reading items from the overflow menu or not, but the `maxSubheaderLinks` seems to be required.

Screenshot 2025-03-18 at 10 28 51 AM
Screenshot 2025-03-18 at 10 29 02 AM

4. I believe this is also unrelated to your changes, but the `ontario` type header is giving errors if no `maxSubheaderLinks` are specified. I think this is meant to be a prop only meant for application headers.

smorrisods (approved):
This looks good to me! There're some failing tests, but otherwise :shipit:

erincarrieresontario:
Hey @MeaghanCarrieres , fixed the first two issues, but the other two issues should be fixed in the parent branch. Will add those as comments when that MR is opened.

PR #35, opened 2025-03-12
feat: added accessibility features to overflow menu

erincarrieresontario:
- Added the ability to navigate through the menu using ‘up/down’ arrow keys.
- Added the functionality to close the menu using the 'esc' key.  
- Added aria-live feature to read out options in the menu.
- The menu button now has aria-expanded="true" when the menu is open and aria-expanded="false" when it is closed.


packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.tsx
```diff
@@ -57,6 +57,16 @@ export class OntarioHeaderApplicationMenu {
 	 */
 	@State() private menuIsOpen: boolean = false;
 
+	/**
+	 * The current index of the menu item that is focused.
+	 */
+	private currentIndex: number = 0;
```
kyle-binger-3, line 63: I think we should include `undefined` as a value and make that the default. Setting it to `0` initially is what causes it to hop to the second item in the list immediately upon pressing the down arrow key (because it's incrementing straight was 0)

Suggested change:
`private currentIndex: number | undefined = undefined;`

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.tsx
```diff
@@ -99,6 +118,70 @@ export class OntarioHeaderApplicationMenu {
...
+	@Listen('keydown', { target: 'window' })
+	handleKeyDown(event: KeyboardEvent) {
+		if (this.menuIsOpen) {
+			const focusableElements = this.menu.querySelectorAll('.ontario-menu-item');
+			const focusableArray = Array.prototype.slice.call(focusableElements);
+
+			switch (event.key) {
+				case 'ArrowDown':
+					event.preventDefault();
+					if (this.currentIndex === -1) {
```
kyle-binger-3, line 136: In your if statement here, you'll also want to check for `undefined` and then set it to 0, this will allow it to increment properly and not jump to the 2nd item on the list right off the hop.

Suggested change:
`if (this.currentIndex === -1 || this.currentIndex === undefined) {`

Maybe add a code comment as well that outlines a scenario where the current index could become `-1`. Is this just a failsafe? Or has it happened in testing?

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.tsx
```diff
@@ -99,6 +118,70 @@ export class OntarioHeaderApplicationMenu {
...
+						this.currentIndex = 0;
+					} else {
+						this.currentIndex = (this.currentIndex + 1) % focusableArray.length;
+					}
+					(focusableArray[this.currentIndex] as HTMLElement).focus();
+					this.updateAriaLive(this.currentIndex);
+					break;
+				case 'ArrowUp':
+					event.preventDefault();
+					if (this.currentIndex === -1) {
```
kyle-binger-3, line 146: See similar note above.

Suggested change:
`if (this.currentIndex === -1 || this.currentIndex === undefined) {`

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.tsx
```diff
@@ -77,13 +87,22 @@ export class OntarioHeaderApplicationMenu {
...
+
 	/**
 	 * This listens for the `menuButtonToggled` event sent from the header menu button when it is clicked.
 	 * It is used to toggle menu visibility by adding or removing the ontario-navigation--open class on the nav element.
 	 */
 	@Listen('menuButtonToggled', { target: 'window' })
 	toggleMenuVisibility(event: CustomEvent<boolean>) {
 		this.menuIsOpen = event.detail;
+		if (this.menuIsOpen) {
+			this.setFocusToFirstMenuItem();
```
kyle-binger-3, line 104: I'm not sure if this is working as intended, it seems if I use the arrow keys to land on the 3rd item of the list, then use escape, tab forwards and then tab backwards, re-open the menu, and press the arrow key down, it will jump to the forth item in the list (meaning the order didn't get reset).

Might be a different way of handling the reset, e.g. on either Tab or Escape we could reset this.currentIndex to undefined.

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.tsx
```diff
@@ -171,6 +254,12 @@ export class OntarioHeaderApplicationMenu {
 						})}
 					</ul>
 				</div>
+				<div
+					id="aria-live-region"
+					class="ontario__visually-hidden"
+					aria-live="polite"
+					ref={(el) => (this.ariaLiveRegion = el as HTMLElement)}
+				></div>
```
kyle-binger-3, line 262: Should these attributes be on the actual `<nav>` element, since it's the element that's being shown or hidden? I'm not an ARIA expert but this just seems like an empty div getting some aria attributes, and that may not be useful to the user.

Haven't tested with a screen reader, just my hunch / suspicion.

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.tsx
kyle-binger-3, line 1: The reason the arrow key focus jumps to the menu button is because of the `endOfMenuReached` event that gets emitted.

```
private linkIsLast() {
    this.endOfMenuReached.emit(true);
}
```

This was originally done to reset the tab order focus on the menu button if the `trapMenuFocus` prop was set to `true`.

Given that Escape (and I think Tab) should now exit the menu, we can probably rethink this functionality and maybe even remove it entirely.

kyle-binger-3 (changes requested):
Hey Erin!

Great job on this! I left some comments regarding the issues that were happening with how the order of elements was out of whack in some cases with the arrow keys.

Also had a general question about the `aria-live-region` div.

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.tsx
erincarrieresontario: This is actually fine practice since although this aria-live-region is visually hidden, it is still accessible to screen readers, so it updates every time the user navigate through the menu items. I tested this with VoiceOver and it works as expected. This is also how we do it for the search-autocomplete!

kyle-binger-3 (approved):
Looks good @erincarrieresontario !

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.scss
```diff
@@ -209,3 +209,14 @@ $navigation-max-width: 72rem;
 		background-color: colours.$ontario-greyscale-5;
 	}
 }
+
+.ontario__visually-hidden {
```
MeaghanCarrieres, line 213: I think it might make sense to add this to the `7-overrides/_visibility.overrides.scss` file in the global styles, to keep all visibility classes together. I also think the naming of the class could just be `ontario-visually-hidden`, not sure if the BEM structure makes sense here.

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.tsx
MeaghanCarrieres, line 1: There's an error with this file when running `pnpm refresh` - I think because you are no longer using Events you'll need to remove both the Event and EventEmitter imports.

smorrisods: I haven't verified this myself, but if it is no longer valid this should be resolved.

packages/ontario-design-system-component-library/src/components/ontario-header/ontario-header.tsx
```diff
@@ -230,6 +230,14 @@ export class OntarioHeader {
 		}
 	}
 
+	@Listen('keydown', { target: 'window' })
+	handleKeyDown(event: KeyboardEvent) {
+		if (this.menuToggle && event.key === 'Escape') {
+			this.menuToggle = false;
```
smorrisods, line 236: This represents a state, rather than an action.  Recommendation, rename `menuToggle` to `menuToggled`.

smorrisods (changes requested):
This overall looks good to me, just two small comments.

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.tsx
erincarrieresontario: It is working now 👍

MeaghanCarrieres: I'm still seeing this error. It's still showing up because both `Event` and `EventEmitter` are being imported at the top of the overflow menu tsx file, but aren't used.

erincarrieresontario: Resolved!

MeaghanCarrieres (approved):

PR #38, opened 2025-04-04
DS-2176 & DS-2177: Ontario Card alt text and heading level props

kyle-binger-3:
This pull request adds two additional props to the `<ontario-card>` component:
 - `headingLevel`
 - `imageAltText`

`headingLevel` is used to pass in a heading level tag (e.g. `<h2>`- `<h6>`) that the card label should render as. It defaults to a value of `h2`.

`imageAltText` is used to add alt text to a card's image. It defaults to an empty string `''`.

There is also a small breaking change in this request:

The `layout` prop has been renamed to `layout-direction`.

**Other improvements**

- State is now managed through a `cardState` object
- warning messages for incorrect prop values are now handled through one consolidated function called `printPropWarning()`
- `HeadingLevelType` has been added as a global/common type as opposed to a component specific type


packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
```diff
@@ -1,15 +1,17 @@
 import { Component, Prop, Element, h, State, Watch } from '@stencil/core';
 import {
-	HeaderColour,
-	HeaderColours,
+	headerColours,
```
erincarrieresontario, line 3: We should be consistent with our capitalization at the beginning for all of our card types.

erincarrieresontario (dismissed):
Just one small comment. Otherwise this looks really good! Tested and works great! Nice work @kyle-binger-3 💯

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
```diff
@@ -127,67 +156,106 @@ export class OntarioCard {
...
 	/**
 	 * Watch for changes to the `headerColour` property for validation purposes.
 	 *
-	 * If the user input doesn't match one of the array values then `headerColour` will be kept empty ('').
+	 * If the user input doesn't match one of the array values then `headerColour` will be kept empty (undefined).
 	 * If a match is found in one of the array values then `headerColour` will be set to the matching array key value.
 	 */
 	@Watch('headerColour')
 	validateHeaderColour() {
 		if (this.headerColour) {
```
matt-stjean, line 201: do we want to explicitly set this to undefined since "this.headColour" could be false for more than one reason?

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
```diff
@@ -250,16 +319,26 @@ export class OntarioCard {
...
+					 *  - The innerHTML such as a string, or additional HTML elements
+					 */}
+					{h(
+						this.cardState.headingLevel, //tag
+						{ className: this.getCardHeadingClasses() }, //attributes
 						<a href={this.getHref()} aria-label={this.ariaLabelText}>
 							{this.label}
-						</a>
-					</h2>
+						</a>,
```
matt-stjean, line 340: this is slick

matt-stjean (dismissed):
looks good to me, but would like others to weigh in too

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
matt-stjean: i did not do a good job referencing the code here, but I hope it makes sense still

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
```diff
@@ -32,13 +34,40 @@ export class OntarioCard {
...
+	 *
+	 * You can find guidance on when to add alt text to an image
+	 * on the Ontario.ca web content editing guide.
+	 *
+	 * https://www.ontario.ca/page/ontario-ca-web-content-editing-guide#alt-text-image-accessibility
+	 *
+	 * Note: This should default to an empty string ('') to ensure the alt attribute appears in the markup
+	 * for decorative images. If left as undefined, the alt attribute will not render in markup.
+	 */
+	@Prop() imageAltText?: string = '';
```
smorrisods, line 69: Being an option type, it might be cleaner to just let it sit as undefined until the user needs it.  I'm not too sure on best practises in TypeScript for this pattern.  Perhaps this is playing into the idea that alt text should be set tp `''` when not describing anything? If this is a case a comment on this line would be handy to make sure we keep this flow in tact if we do any refactoring in the future.

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
```diff
@@ -59,14 +88,14 @@ export class OntarioCard {
 	 * If no type is passed, it will default to 'vertical'.
 	 *
 	 */
-	@Prop() layout?: Layout = 'vertical';
+	@Prop() layout?: LayoutType = 'vertical';
```
smorrisods, line 91: I'll echo my previous comment here about keeping it as the Singular so folks using it externally have nice names to work with. This one might go a step further in it's name and be called `LayoutDirection` since that is what it represents in the way you're using it.  I suppose this might also apply to the property name.

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
```diff
@@ -59,14 +88,14 @@ export class OntarioCard {
...
 
 	/**
 	 * Set the card's header colour.
 	 *
 	 * This is optional.
 	 */
-	@Prop() headerColour?: HeaderColour;
+	@Prop() headerColour?: HeaderColourType;
```
smorrisods, line 98: I'll echo my previous comment here about keeping it as the singular.

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
```diff
@@ -127,67 +156,106 @@ export class OntarioCard {
 	@Watch('layout')
 	validateLayout() {
 		if (this.layout) {
-			const isValid = validateValueAgainstArray(this.layout, Layouts);
-			if (isValid) {
-				this.layoutState = this.layout;
-			} else {
-				this.warnDefaultLayout();
-				this.layoutState = 'vertical';
+			const isValid = validateValueAgainstArray(this.layout, layouts);
```
smorrisods, line 159: We've been using Capitals for these psudeu types that represent the underlying constants of the underlying `type` we're constructing, this breaks from that convention.

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
```diff
@@ -127,67 +156,106 @@ export class OntarioCard {
 	@Watch('layout')
 	validateLayout() {
```
smorrisods, line 157: This might be better named `validateLayoutDirection()`.

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
```diff
@@ -127,67 +156,106 @@ export class OntarioCard {
...
 	validateHeaderColour() {
 		if (this.headerColour) {
-			const isValid = validateValueAgainstArray(this.headerColour, HeaderColours);
+			const isValid = validateValueAgainstArray(this.headerColour, headerColours);
 
 			if (!isValid) {
-				this.warnDefaultHeaderColour();
-				this.headerColour = '';
+				this.printPropWarning('header-colour', '<ontario-card>', this.headerColour, headerColours, 'undefined');
+				this.cardState.headerColour = undefined;
```
smorrisods, line 206: This will most likely not trigger a rerender since the reference that is changing is inside the object.  If this is the intended behaviour here then you'll want to make a copy of the `cardState` object, update the `headerColour` and then set `this.cardState` to the new object to update `this.cardState`'s reference.

_Edit_: I see later on, https://github.com/ongov/ontario-design-system/pull/38/files#diff-a669da9382f7dd4972ce448db598a632c1137c3302ee970b058796e4b5ef3b78R255, you have this covered with a function! Should it be used here?

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
kyle-binger-3: Good question, I wasn't 100% what the right state for this would be. We don't have a default colour, so perhaps null or empty string? But undefined felt more appropriate since it doesn't match any of the options and is essentially being "reset"?

Open to thoughts

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
```diff
@@ -127,67 +156,106 @@ export class OntarioCard {
...
+	 * @param {string} defaultValue - Stringified representation of the value that the corresponding State Object value will default to
 	 */
-	private warnDefaultLayout() {
+	private printPropWarning(
+		propName: string,
+		component: string,
+		propValue: any,
+		acceptableValues: readonly any[],
+		defaultValue: string,
+	) {
```
smorrisods, line 229: This is clever! Good work 😁

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
smorrisods: Seconded 👍🏻👍🏻

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
```diff
@@ -250,16 +319,26 @@ export class OntarioCard {
 				{this.image && (
 					<div class="ontario-card__image-container">
 						<a href={this.getHref()} aria-label={this.ariaLabelText}>
-							<img class="ontario-card__image" src={this.image} />
+							<img class="ontario-card__image" alt={this.imageAltText} src={this.image} />
```
smorrisods, line 322: What happens when `this.imageAltText` is `undefined`? I'm guessing the render process vanished the `alt` prop from the output 🤔

packages/ontario-design-system-component-library/src/utils/common/common.interface.ts
```diff
@@ -95,3 +95,12 @@ export interface Hint extends Base {
...
 export type HintContentType = 'string' | 'html';
+
+export const headingLevels = ['h2', 'h3', 'h4', 'h5', 'h6'] as const;
+
+/**
+ * Create a type from an array of values.
+ *
+ * https://steveholgado.com/typescript-types-from-arrays/
+ */
+export type HeadingLevelType = (typeof headingLevels)[number];
```
smorrisods, line 106: I recommend naming this `HeadingLevel`

smorrisods (changes requested):
Excellent work, @kyle-binger-3, on your updates.  They're solid improvements to the cards.  I've made a few comments to think over and some minor little things I noticed.  If you have any questions/comments add them here or hit me up.

😸

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
kyle-binger-3: Correct! The property will not render at all. According to accessibility guidelines, even if an image has no alt text (say for decorative purposes) it should still render the alt property, just with no value.

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
kyle-binger-3: This is a good catch, thanks!

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
kyle-binger-3: For now I will keep the default as `undefined` but happy to change it in the future if we feel it should be represented some other way.

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
```diff
@@ -32,13 +34,40 @@ export class OntarioCard {
...
+	/**
+	 * The heading level that the label will be rendered as.
+	 *
+	 * @example
+	 *	<ontario-card
+	 *		heading-level="h4"
+	 *		label="Card Title 1"
+	 *	>
+	 */
+	@Prop() headingLevel: HeadingLevelType = 'h2';
```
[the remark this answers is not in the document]
kyle-binger-3: Hey @smorrisods thanks for the great offline discussion about naming conventions. I will drop a high level summary of our notes from that discussion here:

- constants can be lowerCamelCase
- constants can use a suffix (e.g. `headerColourDefinitions`) where `Definitions` is the suffix
- types should be UpperCamelCase
- types **should not** end in `Type` as a suffix (e.g. HeaderColourType)

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
kyle-binger-3: @smorrisods I've updated the Property name as well to `layoutDirection` and marked it as a breaking change (though a minor one).

packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.tsx
kyle-binger-3: Had a discussion about this with Scott as well, here are the notes
https://github.com/ongov/ontario-design-system/pull/38#discussion_r2038329762

smorrisods (dismissed):

smorrisods (approved):

erincarrieresontario (approved):
looks good!

PR #44, opened 2025-04-28
Story/ds 2148/modularize header overflow menu

kyle-binger-3:
This PR brings more modularity to our `ontario-header` component by creating an `ontario-header-overflow-menu` sub-component.

This sub-component has improved accessibility features such as:
 - `Up` and `Down` arrow key navigation
 - Ability to exit with the `Escape` button
 - `aria-live-region` messaging detailing which menu item number you are on in the set (e.g. 1 of 14)

The `ontario-header` now also takes a slot for the application menu, which defaults to the new `ontario-header-overflow-menu` sub-component.

Numerous code improvements have also been made to the `ontario-header` component, including:
 - Large `CSS` refinements and improvements
 - Streamlined markup

This work includes a breaking change to the `SubHeaderLinks` properties on the `ontario-header`. Rather than being separate standalone properties such as: `maxSubheaderDesktopLinks`, `maxSubheaderTabletLinks`, `maxSubheaderMobileLinks`, they are now nested under one object, with each device being a separate property.

e.g.
```
maxSubheaderLinks?: {
    desktop?: number;
    tablet?: number;
    mobile?: number;
};
```


packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/readme.md
```diff
@@ -0,0 +1,88 @@
+import { OntarioHeaderOverflowMenu } from '@ongov/ontario-design-system-component-library-react';
+import useBaseUrl from '@docusaurus/useBaseUrl';
+
+# ontario-header-overflow-menu
+
+This component is intended to be used as a sub-component of the ontario-header component.
+
```
MeaghanCarrieres, line 7: Would it be worth it to link to the ontario-header component here?

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/readme.md
```diff
@@ -0,0 +1,88 @@
...
+  </OntarioHeaderMenuOverflow>
+</div>
+
+<!-- Auto Generated Below -->
+
+## Properties
+
+| Property    | Attribute    | Description                             | Type                   | Default     |
+| ----------- | ------------ | --------------------------------------- | ---------------------- | ----------- |
+| `menuItems` | `menu-items` | The items that will go inside the menu. | `MenuItem[] \| string` | `undefined` |
```
MeaghanCarrieres, line 70: I think in some of the other components (for example, the footer), we have a section for custom property types (like this menu item) where we outline the names, types, etc. I think that would be useful documentation here for the `MenuItem` type.

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.interface.ts
```diff
@@ -0,0 +1,6 @@
+export interface MenuItem {
```
MeaghanCarrieres, line 1: I think this also exists in the ontario-header interface file - is there a way for them to both make use of a single interface?

packages/ontario-design-system-component-library/src/components/ontario-header/ontario-header.tsx
```diff
@@ -358,26 +396,17 @@ export class OntarioHeader {
...
 		return (
 			<li class={liClass}>
-				<a
-					class={linkIsActive === true ? `ontario-link--active` : ``}
-					href={href}
-					onClick={onClick}
-					onBlur={onBlur ? this.trapMenuFocus : undefined}
-					data-type={type}
-				>
+				<a class={linkIsActive === true ? `ontario-link--active` : ``} href={href} onClick={onClick}>
```
MeaghanCarrieres, line 409: This doesn't seem to be working - it doesn't look like the class or styles are being applied.

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.tsx
```diff
@@ -0,0 +1,226 @@
...
+						this.currentIndex = focusableElements.length - 1;
+					} else {
+						this.currentIndex = (this.currentIndex - 1 + focusableElements.length) % focusableElements.length;
+					}
+					focusableElements[this.currentIndex].focus();
+					this.updateAriaLive(this.currentIndex);
+					break;
+				case 'Tab':
+					focusableElements.forEach((element, index) => {
+						element.addEventListener('focus', () => {
```
MeaghanCarrieres, line 134: We discussed this in our demo session yesterday - about looking into whether or not this is adding multiple event listeners!

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.tsx
```diff
@@ -0,0 +1,226 @@
...
+	 *				"linkIsActive": "false"
+	 *			}]'>
+	 *	</ontario-header-overflow-menu>
+	 */
+	@Prop() menuItems: MenuItem[] | string;
+
+	/**
+	 * The `menuItems` are reassigned to `menuItemState` for parsing by `parseMenuItems()`.
+	 */
+	@State() private menuItemState: MenuItem[];
```
MeaghanCarrieres, line 41: Can't remember where we landed on this based on our discussion yesterday, but did we want to do some validation of the JSON in the parseMenuItems function?

smorrisods:
Hey @kyle-binger-3, great work, I'm about to dive in for a review.  I just noticed that the commits are not on develop, possibly due to them being refactored.

I recommend rebasing against `develop` to maintain continuity.

image

packages/ontario-design-system-component-library/src/components/ontario-header/readme.md
```diff
@@ -294,17 +294,25 @@ To ensure best practices, it is important to limit the number of navigation link
 | `menuItems`             | `menu-items`              | The items that will go inside the menu.                                                                                                                                                                                                                | `MenuItem[] \| string`                                        | `undefined`     |
 | `type`                  | `type`                    | The type of header.                                                                                                                                                                                                                                    | `"application" \| "ontario" \| "serviceOntario" \| undefined` | `'application'` |
 
+## Events
+
+| Event               | Description | Type                   |
+| ------------------- | ----------- | ---------------------- |
+| `menuButtonToggled` |             | `CustomEvent<boolean>` |
```
smorrisods, line 301: This event is missing a description.

packages/ontario-design-system-component-library/src/styles/header.scss
```diff
@@ -0,0 +1,97 @@
...
+	z-index: zIndex.$ontario-z-index-above-high;
+
+	@media screen and (min-width: breakpoints.$medium-breakpoint) {
+		box-shadow: 0 globalVariables.$global-radius 12px rgba(0, 0, 0, 0.3);
+		border-radius: globalVariables.$global-radius;
+		display: none;
+		position: absolute;
+		right: spacing.$spacing-4;
+		top: calc(100% - #{spacing.$spacing-2});
+		min-width: 290px; // this value is to control the min-width size the dropdown navigation list can be
```
smorrisods, line 60: I wonder if this should be in `rem`?

packages/ontario-design-system-global-styles/src/styles/scss/7-overrides/_visibility.overrides.scss
```diff
@@ -128,3 +128,14 @@
 		display: block !important;
 	}
 }
+
+.ontario-visually-hidden {
```
smorrisods, line 132: I think he have a class for this already? `show-for-sr`?

smorrisods (changes requested):
It's looking good @kyle-binger-3! I've added some comments and suggestions, let me know if you have any questions/comments.

.github/workflows/main.yml
matt-stjean, line 1: Was this work needed for the overflow menu sub-component work? Just trying to understand the changes.

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.scss
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	a {
+		color: colours.$ontario-colour-black;
+		display: block;
+		font-size: 1.125rem;
+		font-weight: fontWeights.$ontario-font-weights-semi-bold;
+		font-family: fonts.$ontario-font-open-sans;
+		text-decoration: none;
+		padding: calc(spacing.$spacing-3 - math.div(spacing.$spacing-1, 3.25)) spacing.$spacing-5;
```
matt-stjean, line 51: This calculation is a little unclear to me. Can you document what's going on here or pre-calculate it into a new spacing variable.

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.scss
```diff
@@ -0,0 +1,99 @@
...
+
+		&:hover {
+			text-decoration: underline;
+			background-color: $navigation-link-bg--hover;
+		}
+
+		&:active {
+			text-decoration: underline;
+			background-color: colours.$ontario-greyscale-20;
+		}
```
matt-stjean, line 61: `&:active` is being change here as well as below

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.scss
```diff
@@ -0,0 +1,99 @@
...
+
+		&:focus,
+		&:active {
```
matt-stjean, line 64: The second spot where &:active is being changed.

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.scss
```diff
@@ -0,0 +1,99 @@
...
+$navigation-link-bg--hover: #e5e5e5;
+
+.ontario-application-navigation {
+	@extend %ontario-application-navigation;
+
+	ul {
+		list-style: none;
+		padding: spacing.$spacing-2 spacing.$spacing-0;
+		margin: spacing.$spacing-0;
+		border-bottom: 1px solid colours.$ontario-greyscale-20;
```
matt-stjean, line 27: Are we okay with `px`? @smorrisods Or are we supposed to primarily use `rem`?

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.tsx
matt-stjean, line 1: This is really well structured with clear attention to accessibility, state management, and responsiveness. I love it

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.tsx
```diff
@@ -0,0 +1,226 @@
...
+		if (!Array.isArray(this.menuItems) && typeof this.menuItems === 'string') {
+			this.menuItemState = JSON.parse(this.menuItems);
+		} else {
+			this.menuItemState = this.menuItems;
+		}
+
+		// if a menu item is the active page, update the linkIsActive value for that menu item
+		this.menuItemState.map((item) => {
+			const activeLinkRegex = item.href.replace(/\s+/g, '-').toLowerCase();
+			item.linkIsActive = window.location.pathname.includes(activeLinkRegex);
```
matt-stjean, line 66: This might be too loose. We could make the matching a bit stricter

```js
item.linkIsActive = window.location.pathname.toLowerCase() === item.href.toLowerCase();
```

or a little more flexible, but still preventing partial paths:

```js
item.linkIsActive = window.location.pathname.toLowerCase().startsWith(item.href.toLowerCase());
```

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+	render() {
+		return (
+			<nav
+				role="navigation"
+				class={
+					this.menuIsOpen ? 'ontario-application-navigation ontario-navigation--open' : 'ontario-application-navigation'
+				}
+				id="ontario-application-navigation"
+				ref={(el) => (this.menu = el as HTMLElement)}
```
matt-stjean, line 208: Anonymous functions can be sketchy in JSX. It can cause performance issues and cause multiple reloads depending on framework and environment.

This would be safer as a class function
```js
private setMenuRef = (el: HTMLElement | null) => {
	if (el) this.menu = el;
};
...
ref={this.setMenuRef}

File: packages/ontario-design-system-component-library/src/components/ontario-header/service-ontario-header.scss [hunk 16]
```diff
@@ -7,12 +7,15 @@
...
+@use '@ongov/ontario-design-system-global-styles/dist/styles/scss/2-tools/placeholder/focus.placeholders' as
+	placeholders;
 @use '@ongov/ontario-design-system-global-styles/dist/styles/scss/1-variables/typography.variables';
 @use './ontario-application-header.scss';
+@use '../../styles/header.scss';
 
 .ontario-service-header {
 	height: 3.125rem;
-	z-index: calc(zIndex.$ontario-z-index-above-high + 1);
+	z-index: zIndex.$ontario-z-index-above-high;
```
matt-stjean, line 18: Why were we able to remove the +1?

packages/ontario-design-system-component-library/src/styles/header.scss
```diff
@@ -0,0 +1,97 @@
...
+	}
+}
+
+%ontario-overlay {
+	position: fixed;
+	top: 0;
+	left: 0;
+	background: rgba(0, 0, 0, 0.6);
+	width: globalVariables.$full-width;
+	width: globalVariables.$full-width;
```
matt-stjean, line 72: duplicate lines

packages/ontario-design-system-component-library/src/styles/header.scss
matt-stjean, line 60: The rest of the widths are tokenized, could this one be too?

matt-stjean (approved):
Looks great, Kyle

packages/ontario-design-system-component-library/src/components/ontario-header/service-ontario-header.scss
kyle-binger-3: @erincarrieresontario was able to rework the z-index token values, there were some inconsistencies there. `zIndex.$ontario-z-index-above-high` is now rejigged in a way that works so that the `+1` is not needed

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.tsx
kyle-binger-3: This is a good note, thanks! I didn't know that about anon functions in JSX

smorrisods: One thing that could be a problem using anonymous functions is if the `ref` property is being watched and a render triggers a new function to be created each pass then each pass could trigger a rerender causing an infinite loop since the underlying reference changes each time a new one is created.

However, I don't think this is the case with Stencil because their official documentation uses this same pattern we're using.

> ```tsx
> @Component({
>   tag: 'app-home',
> })
> export class AppHome {
> 
>   textInput!: HTMLInputElement;
> 
>   handleSubmit = (event: Event) => {
>     event.preventDefault();
>     console.log(this.textInput.value);
>   }
> 
>   render() {
>     return (
>       <form onSubmit={this.handleSubmit}>
>         <label>
>           Name:
>           <input type="text" ref={(el) => this.textInput = el as HTMLInputElement} />
>         </label>
>         <input type="submit" value="Submit" />
>       </form>
>     );
>   }
> }
> ```
> 
> In this example we are using ref to get a reference to our input `ref={(el) => this.textInput = el as HTMLInputElement}`. We can then use that ref to do things such as grab the value from the text input directly `this.textInput.value`.

Source: https://stenciljs.com/docs/templating-jsx#getting-a-reference-to-a-dom-element

packages/ontario-design-system-component-library/src/components/ontario-header/ontario-header.tsx
```diff
@@ -280,23 +279,67 @@ export class OntarioHeader {
...
 	}
 
+	@Listen('keydown', { target: 'window' })
+	handleKeyDown(event: KeyboardEvent) {
+		if (this.menuToggled && event.key === 'Escape') {
+			this.menuToggled = false;
+			this.menuButtonToggled.emit(this.menuToggled);
+			this.focusMenuButton();
+		}
+	}
```
smorrisods, line 289: `'Escape'` is a magic string here.  I doubt it is going to change but it might be clean to use a lib like [`keycode-js`](https://github.com/kabirbaidhya/keycode-js) or something similar to use a const value.

```ts
import { VALUE_ESCAPE} from 'keycode-js';

console.log('Escape Key:', (e.key === KeyCode.VALUE_ESCAPE));
```

packages/ontario-design-system-component-library/src/components/ontario-header/ontario-header.tsx
```diff
@@ -280,23 +279,67 @@ export class OntarioHeader {
...
+		// screen breakpoint em values
+		const breakpoints = {
+			small: Number(OntarioBreakpointsSmall.replace('em', '')),
+			medium: Number(OntarioBreakpointsMedium.replace('em', '')),
+			large: Number(OntarioBreakpointsLarge.replace('em', '')),
+		};
+
+		const standardFontSize = 16;
+		let windowWidthPx = window.innerWidth;
+		let windowWidthEm = Math.ceil(windowWidthPx / standardFontSize);
```
smorrisods, line 330: These can be `const`s.


```suggestion
		const windowWidthPx = window.innerWidth;
		const windowWidthEm = Math.ceil(windowWidthPx / standardFontSize);
```

packages/ontario-design-system-component-library/src/components/ontario-header/ontario-header.tsx
```diff
@@ -280,23 +279,67 @@ export class OntarioHeader {
...
+
+		let device: DeviceType;
+
+		if (windowWidthEm <= breakpoints.small) {
+			device = 'mobile';
+		} else if (windowWidthEm > breakpoints.small && windowWidthEm <= breakpoints.medium) {
+			device = 'tablet';
+		} else {
+			device = 'desktop';
+		}
```
smorrisods, line 340: This might be helpful as a helper function for other components to use should they need it.  I wonder if it might be a good idea to expose this to our users as well since they could use it in their own components 🤔.

packages/ontario-design-system-component-library/src/components/ontario-header/ontario-header.tsx
```diff
@@ -535,26 +528,27 @@ export class OntarioHeader {
...
-			return false;
-		}
+			if (numOfMenuItems <= 0) {
+				return false;
+			}
 
-		if (viewportSize === 'mobile') {
-			return numOfMenuItems - (maxSubheaderMobileLinks ?? 0) > 0;
-		}
+			if (viewportSize === 'mobile') {
```
smorrisods, line 541: These strings should be codified so we can use a common definition of them.  Perhaps a new type to represent viewport sizes?

packages/ontario-design-system-component-library/src/components/ontario-header/ontario-header.tsx
```diff
@@ -692,47 +693,12 @@ export class OntarioHeader {
...
-													item.onClickHandler,
-													lastLink,
-												);
-									})}
-								</ul>
-							</div>
-						</nav>
+						{/* Minor styling differences in placement of ontario.ca menu and application menu */}
+						<ontario-header-overflow-menu
+							menu-items={JSON.stringify(this.menuItemState)}
```
smorrisods, line 698: Since we're inside the component we should use direct references rather than calling `stringify()` as it adds a cost to serialize then deserialize the data.

packages/ontario-design-system-component-library/src/components/ontario-header/ontario-header.tsx
```diff
@@ -786,138 +752,39 @@ export class OntarioHeader {
...
+
+							<slot name="menu-overflow">
+								<ontario-header-overflow-menu
+									menu-items={JSON.stringify(
+										this.menuItemState?.slice(
+											this.applicationHeaderInfoState?.maxSubheaderLinks?.[this.breakpointDeviceState],
+											this.menuItemState.length,
+										),
+									)}
+								></ontario-header-overflow-menu>
```
smorrisods, line 783: We should avoid, if possible, passing string data between the component directly.  The string `type` is only there to handle the HTML use of the components when there isn't a JS environment to use.

packages/ontario-design-system-component-library/src/styles/header.scss
smorrisods: We should use `rem` to avoid the hardcoded `px` value that won't scale with what the user selects as their font size. `18.125rem` will let the browser do what it needs to do. We also have a `px-to-rem` helper in SASS if we don't want to set the `rem` value directly.

packages/ontario-design-system-component-library/package.json
```diff
@@ -72,6 +72,7 @@
 		"test:unit": "stencil test --spec"
 	},
 	"dependencies": {
+		"@ongov/ontario-design-system-design-tokens": "^5.0.0",
```
smorrisods, line 75: This version number looks to be off from the current `alpha` in this branch, it should be double checked.

smorrisods (changes requested):
I've re-reviewed this PR and added a few comments.  Let me know if there are any questions.

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.scss
kyle-binger-3: This is a good find, I would think we would want our active and focus styles to be idnetical, so why are we setting slightly different styles in the above active state but not applying them to both focus and active here.

I'll investigate further.

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.scss
kyle-binger-3: https://github.com/ongov/ontario-design-system/pull/44#discussion_r2389107081

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.tsx
kyle-binger-3: I like this idea but I think it expands beyond the scope of this work. We have other components that take in json objects as well, I think we can bust this item out as a research ticket.

kyle-binger-3: I've created DS-2365 to look at this separately

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.tsx
```diff
@@ -0,0 +1,226 @@
...
+	 * @param href - the href of the menu item
+	 * @param title - the title of the menu item
+	 * @param linkIsActive - when set to true, this will add the classes necessary to style the link in a way that indicates to the user what the active page/link is
+	 * @param liClass - if there is a class that is related to the <a> portion of the menu item, put it here
+	 * @param onClick - for any custom onClick event a user might want to add to their menu links
+	 */
+	private generateMenuItem(
+		href: string,
+		title: string,
+		linkIsActive: boolean | undefined,
```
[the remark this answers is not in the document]
kyle-binger-3: Done!

packages/ontario-design-system-component-library/src/components/ontario-header/ontario-header.tsx
kyle-binger-3: I think these are less magic than they appear, they are part of the emitted values from KeyBoard events.

- https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key
- https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values

We also check against them within the `ontario-header-overflow-menu` component when navigating the overflow menu with the keyboard:

```
/**
	 * Listens for keydown events and handles navigation within the menu using the ArrowUp and ArrowDown keys.
	 */
	@Listen('keydown', { target: 'window' })
	handleKeyDown(event: KeyboardEvent) {
		if (this.menuIsOpen) {
			const focusableElements: NodeListOf<HTMLElement> = this.menu.querySelectorAll('.ontario-menu-item');

			switch (event.key) {
				case 'ArrowDown':
					event.preventDefault();
					if (this.currentIndex === -1 || this.currentIndex === undefined) {
						this.currentIndex = 0;
					} else {
						this.currentIndex = (this.currentIndex + 1) % focusableElements.length;
					}
					focusableElements[this.currentIndex].focus();
					/**
					 * Tried other ways of triggering this function to condense code via a watch on currentIndex,
					 * and also tried sticking at the bottom of the switch's parent if statement.
					 *
					 * Both attempts led to inconsistent results (e.g. the aria message was out of sync with currentIndex)
					 * or just didn't appear at all. Setting this at the source seems to have the best outcome.
					 */
					this.updateAriaLive(this.currentIndex);
					break;
				case 'ArrowUp':
					event.preventDefault();
					if (this.currentIndex === -1 || this.currentIndex === undefined) {
						this.currentIndex = focusableElements.length - 1;
					} else {
						this.currentIndex = (this.currentIndex - 1 + focusableElements.length) % focusableElements.length;
					}
					focusableElements[this.currentIndex].focus();
					this.updateAriaLive(this.currentIndex);
					break;
				case 'Tab':
					focusableElements.forEach((element, index) => {
						element.addEventListener('focus', () => {
							this.currentIndex = index;
							this.updateAriaLive(this.currentIndex);
						});
					});
					break;
			}
		}
	}
```

One thing I can do is switch `type: any` to `type: KeyboardEvent`.

kyle-binger-3: Ah I was mixing up lines, the code you referenced does include `event: KeyboardEvent` however another section of code does not:

```
/**
	 * The onEscapePressed function clears the searchbar form when Escape is pressed
	 */
	private onEscapePressed(event: any) {
		if (event.key === 'Escape') {
			event.path[0].value = '';
		}
	}
```

I will update this one to be `event: KeyboardEvent`

kyle-binger-3: This turned out to be a small can of worms, leaving it be for now.

packages/ontario-design-system-global-styles/src/styles/scss/7-overrides/_visibility.overrides.scss
kyle-binger-3: Good catch, I've removed this class and switched references to it to use `show-for-sr`

packages/ontario-design-system-component-library/src/components/ontario-header/readme.md
kyle-binger-3: This has been updated with a description, thanks!

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.tsx
kyle-binger-3: This was a bit tricky, as JS doesn't have a built-in method for getting event listeners specific to a particular element, however I was able to lock it down.

```
case 'Tab':
	focusableElements.forEach((element, index) => {
		// within an event listener, the scope of "this" changes.
		// To continue keeping the scope global, set a variable that captures it
		// Without this, calls to global variables/methods/functions will not work
		const self = this;
		if (
			!element.hasAttribute('data-has-focus-listener') ||
			element.getAttribute('data-has-focus-listener') === 'false'
		) {
			element.addEventListener('focus', () => self.updateCurrentIndex(index));
			element.setAttribute('data-has-focus-listener', 'true');
		}
	});
	break;
```

Seems to be working!

packages/ontario-design-system-component-library/src/components/ontario-header/ontario-header.tsx
kyle-binger-3: Done!

packages/ontario-design-system-component-library/src/components/ontario-header/ontario-header.tsx
kyle-binger-3: Done!

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.scss
kyle-binger-3: @erincarrieresontario I think that was a pre-existing calculation that got ported over to the overflow scss, but I have to admit I too am a bit unclear on what this calculation is doing.

Do you know? Can it be simplified?

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.scss
kyle-binger-3: @erincarrieresontario this is another spot where focus/active are referenced, do you think it's safe to remove?

packages/ontario-design-system-component-library/package.json
kyle-binger-3: Version `5.0.0` is all I need for my purposes, but happy to switch it to `^6.0.0` when we officially release it.

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.scss
smorrisods: We don't need to use `calc()` here since the Sass Pre-processor will take care of this calculation at processing time.


```suggestion
		padding: (spacing.$spacing-3 - math.div(spacing.$spacing-1, 3.25)) spacing.$spacing-5;
```

Note, in this I left the brackets due to being explicit in enforcing BEDMAS.

packages/ontario-design-system-component-library/src/utils/common/common.variables.ts
```diff
@@ -1,2 +1,22 @@
...
+
+/**
+ * Numeric screen breakpoint values pulled from the design tokens package.
+ * By default these come in as strings (e.g. '40em'). However
+ * for comparision calculations we need the true numeric value.
+ */
+export const screenBreakpoints = {
+	small: Number(OntarioBreakpointsSmall.replace('em', '')),
+	medium: Number(OntarioBreakpointsMedium.replace('em', '')),
+	large: Number(OntarioBreakpointsLarge.replace('em', '')),
```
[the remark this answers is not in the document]
kyle-binger-3: Possibly regex function to just match number

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.scss
erincarrieresontario: This is going to be changed when my work gets merged in so it's fine to leave as is for now 👍

packages/ontario-design-system-component-library/src/styles/header.scss
kyle-binger-3: This `min-width: 290px` shows up in two palces currently.

However I'm not ready to make it a glboal token just yet as there is more CSS standardizing that we will aim to do in future MRs to this header.

What I have done, is created a `var` in this file to represent this value. (The other file already had a var for it).

I've updated both vars to use the `px-to-rem` global function.

e.g.

`$ontario-navigation-container-min-width: globalFunctions.px-to-rem(290);`

Hopefully the two references can be consolidated in a future improvement.

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.tsx
```diff
@@ -0,0 +1,268 @@
...
+		}
+
+		const activeLinkSet = this.menuItemState.some((menuItem) =>
+			menuItem?.linkIsActive ? menuItem.linkIsActive === true : false,
+		);
+
+		// If no active link is set, try to guess and set the active link based
+		// on if the href is included in the URL.
+		if (!activeLinkSet) {
+			let copyOfMenuItemsState = [...this.menuItemState];
```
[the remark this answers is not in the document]
kyle-binger-3: @smorrisods the line below it is a foreach loop that is modify items within the `copyOfMenuItemsState` array.

Given there is modification to internal parts of the array, wouldn't `let` be more appropriate?

packages/ontario-design-system-component-library/src/components/ontario-header-overflow-menu/ontario-header-overflow-menu.scss
```diff
@@ -0,0 +1,99 @@
...
+		}
+	}
+
+	li {
+		border-top: 1px solid colours.$ontario-greyscale-20;
+		padding: spacing.$spacing-0;
+		margin-bottom: spacing.$spacing-0;
+
+		&:last-of-type {
+			border-bottom: 1px solid colours.$ontario-greyscale-20;
```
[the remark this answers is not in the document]
kyle-binger-3: No constant but I've codified it as a variable within the file. In another pass we can see how it might translate to a constant.

packages/ontario-design-system-component-library/src/styles/header.scss
```diff
@@ -0,0 +1,96 @@
...
+	z-index: zIndex.$ontario-z-index-above-high;
+
+	@media screen and (min-width: breakpoints.$medium-breakpoint) {
+		box-shadow: 0 globalVariables.$global-radius 12px rgba(0, 0, 0, 0.3);
+		border-radius: globalVariables.$global-radius;
+		display: none;
+		position: absolute;
+		right: spacing.$spacing-4;
+		top: calc(100% - #{spacing.$spacing-2});
+		min-width: 18.125rem; // ~290px - this value is to control the min-width size the dropdown navigation list can be
```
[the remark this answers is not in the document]
kyle-binger-3: Done!

packages/ontario-design-system-component-library/src/styles/header.scss
```diff
@@ -0,0 +1,96 @@
...
+%ontario-application-navigation__container {
+	background: colours.$ontario-colour-white;
+	position: absolute;
+	transform: translateY(-100%);
+	width: 100%;
+	visibility: hidden;
+	z-index: zIndex.$ontario-z-index-above-high;
+
+	@media screen and (min-width: breakpoints.$medium-breakpoint) {
+		box-shadow: 0 globalVariables.$global-radius 12px rgba(0, 0, 0, 0.3);
```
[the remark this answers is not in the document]
kyle-binger-3: We can review constants at a later date, for now I've created this as an in file variable

packages/ontario-design-system-component-library/src/styles/header.scss
```diff
@@ -0,0 +1,96 @@
...
+	height: globalVariables.$standard-input-height;
+	line-height: 1.5;
+	text-align: center;
+	text-decoration: none;
+}
+
+%ontario-header-button--with-outline {
+	border: globalVariables.$border-size-standard solid colours.$ontario-colour-white;
+	margin: spacing.$spacing-0 spacing.$spacing-0 spacing.$spacing-0 spacing.$spacing-4;
+	padding: 0.375rem spacing.$spacing-4;
```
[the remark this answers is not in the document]
kyle-binger-3: We can review constants at a later date, for now I've created this as an in file variable

erincarrieresontario (approved):
looks great Kyle!

PR #49, opened 2025-05-16
Refactor(angular-poc): Use ontario-form-container component in Angular POC

MeaghanCarrieres:
This MR updates the Angular POC by wrapping applicable form elements in the new `ontario-form-container` component. This helps standardize spacing and visual layout across grouped form controls.

While reviewing the React POC, I decided to not apply the `ontario-form-container`. The React examples are singular examples of components and would benefit more from targeted margin override utility classes (e.g., `.ontario-margin-bottom-16-!)`, which provide better control for layout adjustments in that context.

During this implementation, it became clear that our existing global override classes (for spacing and visibility) don’t consistently apply inside our web components due to Shadow DOM encapsulation. Although these utility classes include the `!important` tag, they don't seem to be taking effect unless `display: block` is also applied on the component itself. This discovery led to the creation of a [follow-up ticket (DS-2232)](https://ontariodigital.atlassian.net/browse/DS-2232) to address how we might ensure that utility classes like margin and padding overrides can be used directly on our web components.

### DIFF
--- a/packages/app-angular/src/app/pages/framefive/framefive.component.html
+++ b/packages/app-angular/src/app/pages/framefive/framefive.component.html
@@ -8,18 +8,22 @@
 
 <div>
 	<form (valueChanges)="saveToTemporaryStorage()">
-		<ontario-textarea
-			name="additionalDetails"
-			hintText="{{ 'form.questions.additionalDetails.hint' | translate }}"
-			[caption]="{
-				captionText: getTranslation().additionalDetailsLabel,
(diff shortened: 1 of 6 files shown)


erincarrieresontario (approved):
Looks good to me!

smorrisods (approved):
LGTM, thanks for making the Jira issue to track it.  Also, the title currently includes the word _DRAFT_, this should be rectified prior to merging.

PR #62, opened 2025-06-03
DS-2244: removed row-gap from card collection to remove extra space

erincarrieresontario:
- Replaced the generic gap property with column-gap and row-gap for more precise control.
- Result: Cards now have consistent horizontal spacing (left and right), with no vertical gaps between rows—creating a tighter, cleaner layout.


smorrisods (approved):
LGTM!

PR #106, opened 2025-07-21
DS-2298 - Create test page for ontario-language-toggle

erincarrieresontario:
This MR adds an ontario-language-toggle to the Next PoC. It includes variants for language, size, url, and customLanguageToggle prop.


packages/app-nextjs/src/app/components/ontario-language-toggle/page.tsx
```diff
@@ -0,0 +1,36 @@
+import { Grid } from '../../grid';
```
MeaghanCarrieres, line 1: This will need a `use client` here for the handleCustomToggle function to work

PR #107, opened 2025-07-25
DS-2272: Extract Docusaurus to its own repo

smorrisods:
This PR extracts the Docusaurus documentation app to its own repository.  This allows us to manage the documentation across all package versions and enables a cleaner design for managing documentatnon.

Some notable updates are:

- Refactored the documentation build and sync process to use artifact-based delivery.
- Docs are now built in the main build workflow and uploaded as a compressed artifact (`docs-artifacts-name`), ensuring a single source of truth for artifact names.
- The Docusaurus sync workflow downloads the docs artifact and updates the external Docusaurus repo, preserving required files and automating the PR process.
- Removed the legacy `app-docs` workflow; all docs build and packaging now occur in the main build pipeline.
- Updated all relevant workflow and package scripts to use the new artifact naming convention.


.github/workflows/build.yml
```diff
@@ -50,3 +60,12 @@ jobs:
...
             !**/node_modules/**/*
+            !packages/docusaurus/**/*
+
+      # Upload docs artifact
+      - name: Upload docs artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ inputs.docs-artifacts-name }}
+          compression-level: 9 # Best compression
+          path: packages/docusaurus/docs/
```
MeaghanCarrieres, line 71: Would it make sense to set artifact retention here so old docs bundles don't pile up?

.github/workflows/sync-docs-to-docusaurus.yml
```diff
@@ -0,0 +1,92 @@
...
+          echo "version=${tag}" >> $GITHUB_OUTPUT
+
+      - name: Update package.json in Docusaurus
+        working-directory: packages/docusaurus
+        run: pnpm add @ongov/ontario-design-system-component-library-react@${{ steps.get_tag.outputs.version }}
+
+      - name: Version Docusaurus docs
+        working-directory: packages/docusaurus
+        run: npx docusaurus docs:version ${{ steps.get_tag.outputs.version }}
+
```
MeaghanCarrieres, line 81: Would it be helpful here to add a step to make sure the `docs-update` label exists in the external repo to avoid label errors? Or is that overkill?

MeaghanCarrieres:
Hey @smorrisods. Nice work. Pulling the docs into their own repo and shipping them as a single build artifact makes our setup feel much cleaner and easier to understand. 👍

I had a few questions (mostly just to help me understand):
1. **Version source:** The workflow grabs the “latest tag”. Will tags always be available in CI so this never ends up blank?
2. **When it runs:** It runs on `master/alpha/beta`. Do we also want this to run on release tags (e.g., v1.2.3) or only on those branches?
3. **Reruns:** If the same version runs again, does Docusaurus handle that, or do we need to skip versioning if it already exists?
4. **Docs cleanup:** The step that wipes docs keeps `_category_.json`. Are there any other files we need to keep so we don’t accidentally delete something important?


Apart from that, how should we go about testing this? Is there a link to a PR that the workflow opened in the external docs repo for a test version?

erincarrieresontario (approved):
LGTM, just added a few suggestions :)

.github/workflows/sync-docs-to-docusaurus.yml
```diff
@@ -0,0 +1,92 @@
...
+          path: packages/docusaurus/docs/
+
+      # Since this follows the build job, the latest tag is the latest version of npm packages.
+      - name: Get latest tag
+        id: get_tag
+        working-directory: .
+        run: |
+          tag=$(git describe --tags --abbrev=0)
+          echo "version=${tag}" >> $GITHUB_OUTPUT
+
```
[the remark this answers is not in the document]
smorrisods: The tags are automatically generated by our release tooling, so there is no validation because this is a consistent format.

.github/workflows/sync-docs-to-docusaurus.yml
smorrisods: This label was manually created in the Docusaurus project.  I tried to do some research into how the GitHub REST API (which is used by this Action under the hood) handles the case of no label but none of their documentation addressed this case.  I'm going to skip validation for now as it appears this edge case isn't an issue, but we can monitor it to see what happens.

.github/workflows/build.yml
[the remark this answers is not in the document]
smorrisods: If any of the previous steps fail then the whole job will be cancelled.  I think this is an acceptable risk because unless something catastrophic happens and it fails without failing the worst that will happen is we'll get an empty PR in the Docusaurus repo

PR #125, opened 2025-09-05
DS-2276: Add ontario footer test page

matt-stjean:
# This PR introduces dedicated visual regression (VR) test pages for the <OntarioFooter> component.

This PR approaches this with three clear, layout-specific pages. Each page now maps directly to the supported type prop values:

- Default Footer → ontario-footer/page-default/page.tsx
- Two-Column Footer → ontario-footer/page-two-column/page.tsx
- Three-Column Footer → ontario-footer/page-three-column/page.tsx

## What's included:

- New VR test pages for default, twoColumn, and threeColumn footer layouts.

- Each page includes baseline examples from official documentation.

- Variants added per page to cover key props:

  - lang (fr, en)

  - footerLinks (full, partial, fallback)

  - socialLinks (expanded layouts only)

  - topMargin (true/false)

  - assetBasePath variations

### DIFF
--- a/packages/app-nextjs/src/app/components/ontario-footer/page-three-column/page.tsx
+++ b/packages/app-nextjs/src/app/components/ontario-footer/page-three-column/page.tsx
@@ -1,6 +1,38 @@
 import { Grid } from '../../../grid';
 import { OntarioFooter } from '@ongov/ontario-design-system-component-library-react';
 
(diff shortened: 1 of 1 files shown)


smorrisods:
@matt-stjean, I'm adding the `next-poc` label.

PR #129, opened 2025-09-12
DS-2154: Update handleHeaderLanguageToggled event to be consistent

erincarrieresontario:
Updated the `handleHeaderLanguageToggled` event to be consistent across the board.


kyle-binger-3 (changes requested):
Overall looks good @erincarrieresontario just a small note about imports

packages/ontario-design-system-component-library/src/components/ontario-loading-indicator/ontario-loading-indicator.tsx
```diff
@@ -61,9 +61,13 @@ export class OntarioLoadingIndicator {
 		}
 	}
 
+	/**
+	 * Handles an update to the language should the user request a language update from the language toggle.
+	 * @param event The language that has been selected.
+	 */
```
kyle-binger-3: There's a few comments like these which are great.

For the `@param` line, could you update it to say `@param {CustomEvent} - The language that has been selected.`

Makes the comment a hair more descriptive to what the actual Type of the param is.

kyle-binger-3 (changes requested):

MeaghanCarrieres (approved):

PR #132, opened 2025-09-15
DS-2233: Fix broken `isOpen` prop on ontario-accordion component

MeaghanCarrieres:
This MR updates the `ontario-accordion` component to simplify state handling and improve clarity. Specifically, it removes the global `isOpen` prop and ensures that only the per-accordion `isOpen` property inside accordionData determines which accordions are initially open.

## Changes
- Removed global `isOpen` prop
   - Initial open state is seeded only from `accordionData[i].isOpen`
- Added new `seedOpenIndexesFromItems()` function
   - Seeds `openAccordionIndexes` directly from single accordion `isOpen` flags.
- Updated `parseAccordionData()` Watch function
   - Normalizes input and applies default values for a`ccordionContentType` and `isOpen`.
   - Calls `seedOpenIndexesFromItems()` to establish initial open state.
   - Improves error handling and consistency when parsing JSON input.
- Updated class and aria handling
- **Expand/Collapse all button**: Logic remains the same, but now correctly reflects state seeded from per-accordion `isOpen`.
- Updated class and aria handling
    - When open:
        - accordion wrapper `<div>` gets `open` class added
        - accordion heading gets `ontario-accordion__heading ontario-expander--active` class added
        - accordion button aria-expanded value is set to `true`
        - accordion content `<section>` gets `ontario-accordion__content ontario-expander__content--opened` class added
    - When closed: correct “closed” classes/attributes are applied
-  Expand/Collapse all button
   - Still toggles all panels, but logic now integrates with per-accordion `isOpen` state
   - Updates labels and aria-expanded consistently
- Utility cleanup
   - Improved readability of seeded indexes logic
   - Maintains unique id generation via new `uidPrefix` state

## Testing
- Verified that items marked `isOpen: true` in `accordionData` render expanded initially.
- Verified that toggling an individual accordion updates classes/aria attributes correctly.
- Verified that Expand/Collapse all reflects current state and toggles all items as expected.

### DIFF
--- a/packages/ontario-design-system-component-library/src/components/ontario-accordion/accordion.interface.ts
+++ b/packages/ontario-design-system-component-library/src/components/ontario-accordion/accordion.interface.ts
@@ -30,6 +30,12 @@ export interface Accordion extends Base {
 	ariaLabelText: string;
 }
 
+export enum AccordionChangeDetailReasons {
+	Init = 'init',
+	ToggleOne = 'toggle-one',
+	ToggleAll = 'toggle-all',
+}
+
 /**
  * The event detail payload emitted by the accordion component.
  */
@@ -44,5 +50,5 @@ export interface AccordionChangeDetail {
 	isBulk?: boolean;
 
 	/** Optional description of what triggered the event */
-	reason?: 'init' | 'toggle-one' | 'toggle-all';
+	reason?: AccordionChangeDetailReasons;
 }
--- a/packages/ontario-design-system-component-library/src/components/ontario-accordion/ontario-accordion.tsx
+++ b/packages/ontario-design-system-component-library/src/components/ontario-accordion/ontario-accordion.tsx
@@ -1,6 +1,7 @@
 import { Component, Prop, Element, Event, EventEmitter, State, h, Listen, Watch } from '@stencil/core';
+import { v4 as uuid } from 'uuid';
 
-import { Accordion, AccordionChangeDetail } from './accordion.interface';
+import { Accordion, AccordionChangeDetail, AccordionChangeDetailReasons } from './accordion.interface';
 import { ExpandCollapseButtonDetails } from './expandCollapseButtonDetails.interface';
 
 import { Language } from '../../utils/common/language-types';
@@ -99,7 +100,7 @@ export class OntarioAccordion {
 	@State() private openAccordionIndexes: number[] = [];
 
 	/** Unique prefix for a11y ids */
-	@State() private uidPrefix: string = `ontario-accordion-${Math.random().toString(36).slice(2, 9)}`;
+	@State() private uidPrefix: string = `ontario-accordion-${uuid()}`;
 
 	/**
 	 * This listens for the `setAppLanguage` event sent from the test language toggler when it is is connected to the DOM. It is used for the initial language when the input component loads.
@@ -129,13 +130,15 @@ export class OntarioAccordion {
 			const parsed = Array.isArray(this.accordionData) ? this.accordionData : JSON.parse(this.accordionData || '[]');
 
 			// Normalize and type-guard
-			this.internalAccordionData = (parsed || []).map((prop: any) => ({
-				label: prop?.label ?? '',
-				content: prop?.content ?? '',
-				accordionContentType: (prop?.accordionContentType ?? 'string') as 'string' | 'html',
-				isOpen: Boolean(prop?.isOpen),
-				ariaLabelText: prop?.ariaLabelText,
-			}));
+			this.internalAccordionData = parsed.length
+				? parsed.map((prop: Accordion) => ({
+						label: prop?.label ?? '',
+						content: prop?.content ?? '',
+						accordionContentType: (prop?.accordionContentType ?? 'string') as 'string' | 'html',
+						isOpen: Boolean(prop?.isOpen),
+						ariaLabelText: prop?.ariaLabelText,
+					}))
+				: [];
 
 			// Seed the openAccordionIndexes based on the isOpen properties
 			this.seedOpenIndexesFromItems();
@@ -145,7 +148,7 @@ export class OntarioAccordion {
 
 			this.emitAccordionChange({
 				openIndexes: this.openAccordionIndexes,
-				reason: 'init',
+				reason: AccordionChangeDetailReasons.Init,
 			});
 		} catch {
 			const message = new ConsoleMessageClass();
@@ -163,7 +166,7 @@ export class OntarioAccordion {
 
 			this.emitAccordionChange({
 				openIndexes: this.openAccordionIndexes,
-				reason: 'init',
+				reason: AccordionChangeDetailReasons.Init,
 			});
 		}
 	}
@@ -229,7 +232,7 @@ export class OntarioAccordion {
 		this.emitAccordionChange({
 			openIndexes: this.openAccordionIndexes,
 			changedIndex: index,
-			reason: 'toggle-one',
+			reason: AccordionChangeDetailReasons.ToggleOne,
 		});
 	}
 
@@ -250,7 +253,7 @@ export class OntarioAccordion {
 		this.emitAccordionChange({
 			openIndexes: this.openAccordionIndexes,
 			isBulk: true,
-			reason: 'toggle-all',
+			reason: AccordionChangeDetailReasons.ToggleAll,
 		});
 	}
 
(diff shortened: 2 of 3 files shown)


packages/ontario-design-system-component-library/src/components/ontario-accordion/ontario-accordion.tsx
```diff
@@ -112,17 +99,47 @@ export class OntarioAccordion {
...
+			this.language = validateLanguage(event);
+		}
+	}
+
+	/**
+	 * Handle the language being toggled from the `<ontario-header>`.
+	 * @param event Event object passed when the event is fired.
+	 */
+	@Listen('headerLanguageToggled', { target: 'window' })
+	handleHeaderLanguageToggled(event: CustomEvent<Language>) {
```
smorrisods, line 120: The type we want to use here for the event is `HeaderLanguageToggleEventDetails` rather than language. See [the definition](https://github.com/ongov/ontario-design-system/blob/77128f51988b90247bd783cf0c8e8f91bec2734d/packages/ontario-design-system-component-library/src/components/ontario-language-toggle/ontario-language-toggle.tsx#L87) for more details.

smorrisods (changes requested):
Looking good, only one minor suggestion to clean up the one event.

packages/ontario-design-system-component-library/src/components/ontario-accordion/ontario-accordion.tsx
```diff
@@ -45,46 +49,33 @@ export class OntarioAccordion {
 	 *
 	 * The `content` is expecting a string, that can either be written as HTML or a just a plain string, depending on the accordionContentType.
 	 *
+	 * - label: string
+	 * - content: string (rendered as text or via innerHTML)
+	 * - accordionContentType: 'string' | 'html'
+	 * - isOpen: boolean (initial open state - default is false)
+	 * - ariaLabelText: string
```
smorrisods: This looks like it could be a cause for tech debt, as it is a string version of the definition of `accordionContentType`.  Perhaps we can use the [`@see` ](https://jsdoc.app/tags-see) tag to reference the type rather than duplicating its properties here?

packages/ontario-design-system-component-library/src/components/ontario-accordion/ontario-accordion.tsx
```diff
@@ -45,46 +49,33 @@ export class OntarioAccordion {
...
-
-	/**
-	 * Handle the language being toggled from the `<ontario-header>`.
-	 * @param event Event object passed when the event is fired.
-	 */
-	@Listen('headerLanguageToggled', { target: 'window' })
-	handleHeaderLanguageToggled(event: CustomEvent<Language>) {
-		this.language = validateLanguage(event);
-	}
+	@Event() accordionChange: EventEmitter<{ openIndexes: number[] }>;
```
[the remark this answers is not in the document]
MeaghanCarrieres: Thanks for catching that @smorrisods! I’ve now implemented the `accordionChange.emit()` logic so the event fires consistently whenever the accordion’s open/closed state changes.

### Updates include:
- Added a helper method `emitAccordionChange(detail: AccordionChangeDetail)` to centralize event emits and keep the logic consistent.
- The event detail type is defined in `accordion.interface.ts` as `AccordionChangeDetail`, which now includes:
```
{
  openIndexes: number[];
  changedIndex?: number;
  isBulk?: boolean;
  reason?: 'init' | 'toggle-one' | 'toggle-all';
}
```
- The event now emits:
   - Once after initialization or data parsing (reason: 'init')
   - After single-item toggles (changedIndex + reason: 'toggle-one')
   - After expand/collapse all actions (isBulk: true, reason: 'toggle-all')

This provides full context to consumers (which indexes are open and what triggered the change) without requiring them to guess which one was toggled. Does this look good or maybe overkill?

packages/ontario-design-system-component-library/src/components/ontario-accordion/ontario-accordion.tsx
```diff
@@ -112,20 +98,55 @@ export class OntarioAccordion {
 	 */
 	@State() private openAccordionIndexes: number[] = [];
 
+	/** Unique prefix for a11y ids */
+	@State() private uidPrefix: string = `ontario-accordion-${Math.random().toString(36).slice(2, 9)}`;
```
kyle-binger-3: We have a `uuid` package installed in this repo, I think we should leverage it here instead.

Screenshot 2025-10-28 at 11 16 50 AM

https://www.npmjs.com/package/uuid

packages/ontario-design-system-component-library/src/components/ontario-accordion/accordion.interface.ts
```diff
@@ -29,3 +29,20 @@ export interface Accordion extends Base {
...
+	openIndexes: number[];
+
+	/** The index that was just toggled (if applicable) */
+	changedIndex?: number;
+
+	/** True if triggered by expand/collapse all */
+	isBulk?: boolean;
+
+	/** Optional description of what triggered the event */
+	reason?: 'init' | 'toggle-one' | 'toggle-all';
```
kyle-binger-3: @smorrisods has encouraged me to break strings like this out into an enum.

e.g.

```
export enum AccordionChangeDetailReasons {
	Init = 'init',
	ToggleOne = 'toggle-one',
	ToggleAll = 'toggle-all',
}
```

In your code, you can now reference these rather than the direct strings.

e.g. reference `AccordionChangeDetailReasons.Init` rather than `init`.

If you need to use these values as a Type, you can create them like this:

```
// Create a Union Type from Enum Values
export type AccordionChangeDetailReason = `${AccordionChangeDetailReasons}`;
```

You can now pass this Type Definition into your AccordionChangeDetail reason field.

e.g.

```
/**
 * The event detail payload emitted by the accordion component.
 */
export interface AccordionChangeDetail {
	/** Array of indexes currently open */
	openIndexes: number[];

	/** The index that was just toggled (if applicable) */
	changedIndex?: number;

	/** True if triggered by expand/collapse all */
	isBulk?: boolean;

	/** Optional description of what triggered the event */
	reason?: AccordionChangeDetailReason;
}
```

packages/ontario-design-system-component-library/src/components/ontario-accordion/readme.md
```diff
@@ -86,13 +86,18 @@ These language change events only fire in the browser after hydration. To ensure
...
+| `language`             | `language`               | The language of the component. This is used for translations, and is by default set through event listeners checking for a language property from the header. If none are passed, it will default to English.                                           | `"en" \| "fr" \| undefined`                          | `undefined` |
+| `name`                 | `name`                   | The name of the accordion component. This is not optional.                                                                                                                                                                                              | `string`                                             | `undefined` |
+
+## Events
+
+| Event             | Description                    | Type                                 |
+| ----------------- | ------------------------------ | ------------------------------------ |
+| `accordionChange` | Emits when open indexes change | `CustomEvent<AccordionChangeDetail>` |
```
smorrisods: The readme should be updated with information about the `AccordionChangeDetail` type.

smorrisods (changes requested):

packages/ontario-design-system-component-library/src/components/ontario-accordion/ontario-accordion.tsx
```diff
@@ -112,20 +98,55 @@ export class OntarioAccordion {
...
+
+	/**
+	 * Handles an update to the language should the user request a language update from the language toggle.
+	 * @param event The language that has been selected.
+	 */
+	@Listen('headerLanguageToggled', { target: 'window' })
+	handleHeaderLanguageToggled(event: CustomEvent<HeaderLanguageToggleEventDetails>) {
+		this.language = validateLanguage(event.detail.newLanguage);
+	}
+
```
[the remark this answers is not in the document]
kyle-binger-3: @MeaghanCarrieres _I think_ line 108 needs to also use `HeaderLanguageToggleEventDetails` as opposed to `Language`.

packages/ontario-design-system-component-library/src/components/ontario-accordion/accordion.interface.ts
MeaghanCarrieres: Good suggestion - updated!

smorrisods (approved):
LGTM 🚀

PR #160, opened 2025-12-23
Alpha -> Develop

smorrisods:
This PR bring `develop` up-to-date with the most recent `alpha`.


erincarrieresontario (approved):
LGTM, but looks like some pipelines are failing @smorrisods

kyle-binger-3 (approved):
LGTM! 🚢

smorrisods:
> LGTM, but looks like some pipelines are failing @smorrisods

The failure was due to no new `alpha` version since there were no `feat` or `fix` commits so the push to make the docs failed.  I need to update that step to check the version field and not run the job if there is no version number to release 🤓! cc/ @erincarrieresontario

PR #161, opened 2025-12-24
DS-1659: Only render application header link if link exists

erincarrieresontario:
- Fix for application header link appearing even if no link is added


smorrisods (approved):
LGTM 👩‍🎤🎸🚀

MeaghanCarrieres (approved):

PR #172, opened 2026-02-02
DS-2239: updated comments to properly reflect the disableDynamicMenu prop

erincarrieresontario:
- Added `handleDisableDynamicMenuChange()` method that resets the `isDynamicMenu` flag when the prop changes. This allows the component to respond to dynamic prop updates (e.g., in React/NextJS applications). 
- Improved Prop documentation


packages/ontario-design-system-component-library/src/components/ontario-header/ontario-header.tsx
```diff
@@ -358,6 +367,17 @@ export class OntarioHeader {
 		}
 	}
 
+	/**
+	 * Watch for changes to the disableDynamicMenu prop to reset the fetch state.
+	 * This allows the menu to be fetched again if the prop is changed after initial load.
+	 */
+	@Watch('disableDynamicMenu')
+	private handleDisableDynamicMenuChange() {
```
smorrisods: There is an error here.  This isn't being called anywhere, which makes sense since it responds to a prop change, so a `// @ts-ignore` with a message should clear this up.

This function also assumes `disableDynamicMenu` is always `true` but what happens if `disableDynamicMenu` is set to false to re-enable the dynamic menu?

We can quickly [get the new and old values](https://stenciljs.com/docs/reactive-data#the-watch-decorator-watch) just from the function call itself.

```suggestion
	private handleDisableDynamicMenuChange(newValue: boolean, oldValue: boolean) {
```

This is the error from the [CI Build Job](https://github.com/ongov/ontario-design-system/actions/runs/21598201187/job/62367070801?pr=172#step:7:134):

```
 [ ERROR ]  TypeScript: src/components/ontario-header/ontario-header.tsx:375:10
             'handleDisableDynamicMenuChange' is declared but its value is never
             read.
  
      L374:  @Watch('disableDynamicMenu')
      L375:  private handleDisableDynamicMenuChange() {
      L376:  	// Reset the dynamic menu flag when the prop changes
```

smorrisods (changes requested):

smorrisods (changes requested):

smorrisods (approved):
LGTM 🚀

PR #181, opened 2026-02-10
DS-2566: Add web component to HTML generator

erincarrieresontario:
# `@ongov/ontario-design-system-component-html-generator`

Introduces **`@ongov/ontario-design-system-component-html-generator`**, a new internal package that renders Ontario Design System Stencil components to static HTML samples.

## What it does

- Hydrates each registered ODS component using Stencil’s server‑side renderer and extracts clean, flattened markup with shadow DOM inlined
- Scopes and transforms component styles so output is self‑contained and portable
- Writes one `.html` file per component to a configurable output directory

### Library
`generateSamples(config)` returns a typed `SampleGeneratorResult` for programmatic use

### CLI
- `generate` command with flags for:
  - Output directory
  - Style inclusion
  - `--full-document`
- `init` subcommand to scaffold a config file

## `--full-document` mode

When passed, output is wrapped in a complete document with:
- charset + viewport meta
- inline `<style>`
- `<html>`, `<head>`, and `<body>` tags

This makes any generated sample immediately previewable in a browser without additional setup.

## Docs & testing

- Jest unit-test suite covering all core HTML/CSS processing utilities
- Package README is automatically copied to the Docusaurus internal docs site as part of the build via the shared `documentation-helper` script


packages/app-sample-generator/package.json
```diff
@@ -0,0 +1,31 @@
...
+		"screenshot:sample-button": "node scripts/capture-sample-button.mjs",
+		"test": "echo \"Error: no test specified\" && exit 1"
+	},
+	"keywords": [
+		"stencil",
+		"docs",
+		"samples"
+	],
+	"author": "",
+	"license": "ISC",
```
smorrisods: We should update the licence to match out APACHE-2.0 licence

packages/app-sample-generator/package.json
```diff
@@ -0,0 +1,31 @@
+{
```
smorrisods: Has this been designed to release as a tool as a package? if so we should update our namespace to include `@ongov`.  If no, then we should park it as private.

smorrisods (commented):
Some quick observations 👀

packages/app-sample-generator/src/cli.ts
```diff
@@ -0,0 +1,38 @@
...
+import path from 'path';
+import { generateSamples } from './index.js';
+import { samples } from './sample-config.js';
+
+// Basic CLI argument parsing
+const args = process.argv.slice(2);
+const options: any = {};
+
+for (let i = 0; i < args.length; i++) {
+	if (args[i] === '--outputDirectory' && args[i + 1]) {
```
smorrisods: I recommend using `commander` for building the CLI tool.

smorrisods: It would be handy to take in a samples file path, and default to a known one if no path is provided.  This avoids the need to conform exactly to a hardcoded path.

smorrisods: Another random idea is to have a sample generator that users can use to generate an sample file from the CLI that they can play with and expand upon.

packages/ontario-design-system-component-html-generator/src/sample-config.ts
smorrisods, line 1: Allowing for users to pass in JSON as well would make on disk configuration really simple.  Everything here appears to be strings or booleans so structurally it isn't overly complected to implement.

smorrisods: I'm assuming this could be an input format for the CLI.

packages/app-sample-generator/src/index.ts
```diff
@@ -0,0 +1,334 @@
...
+ */
+async function renderSample(
+	sample: ComponentSample,
+	renderer: SampleRenderer,
+	styleLoader: SampleStyleLoader,
+): Promise<RenderedSampleParts> {
+	const hydratedHtml = await renderer(sample.html);
+
+	const renderedMarkup = stripHydrationArtifacts(flattenShadowRoot(hydratedHtml));
+	const markup = formatHtml(renderedMarkup, true, true);
```
smorrisods: This fixes the `stripOuterComponent` and `removeStyles` values and doesn't pass configuration option to the formatter.  I recommend we add an option parameter (and expose it to the outside API/CLI.)

packages/app-sample-generator/utils/tests/mutation-observer.mock.ts
smorrisods, line 1: Is this still in use?

packages/app-sample-generator/LICENSE
smorrisods, line 1: All our other work is `Apache 2.0` licenced, we should be consistent.

packages/app-sample-generator/package.json
```diff
@@ -0,0 +1,46 @@
...
+	"name": "@ongov/app-sample-generator",
```
smorrisods: We should brainstorm some names for this and move it into it's own public package.

Something like:

`@ongov/ontario-design-system-component-html-generator`

packages/ontario-design-system-component-html-generator/package.json
```diff
@@ -0,0 +1,31 @@
+{
+	"name": "stencil-sample-generator",
+	"version": "1.0.0",
+	"description": "Automated HTML sample generator for Stencil components.",
+	"main": "index.js",
+	"scripts": {
```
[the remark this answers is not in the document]
smorrisods: Once we move it to being it's own library we can explicitly set `"private": false,`

smorrisods (changes requested):
@erincarrieresontario, this is amazing! Awesome work. 😁 

I've left some questions, comments, and ideas in my review here.

One thing we should add before it is ready is a set of tests to keep it honest.  There are plenty of core library functions, like `formatHtml`. `stripHydrationArtifacts`, etc. that can be tested with unit tests and added to our CI/CD pipeline.

packages/app-sample-generator/utils/tests/mutation-observer.mock.ts
erincarrieresontario: Nope - good catch!

erincarrieresontario:
Hey @smorrisods , since you last reviewed this, here are the changes: 

**Main change:** the old app-sample-generator package was replaced/restructured as ontario-design-system-component-html-generator.
**Renames/moves:** most existing package files were moved to the new package path (README, configs, sample outputs, type files, etc.).
**New additions:** added fresh LICENSE, jest config, new index.ts, cli.ts, utils.ts, and utils.test.ts in the new package.
**Removals from old package**: old index.ts,  cli.ts, old license, and one old test mock file were deleted from the previous package path.
**Dependency metadata:** package.json and pnpm-lock.yaml were updated to reflect the new package structure/name.

packages/ontario-design-system-component-html-generator/generated-samples-refactor-check-2/ontario-accordion.html
smorrisods, line 1: I just realized this isn't fully formed semantic HTML and it's missing the DOCTYPE and `<html>` tags.  I think we should have an option to add this for those that want to preview the pages.

packages/ontario-design-system-component-html-generator/README.md
smorrisods, line 1: I just realized we'll want to add the plumbing so this gets copied to the docs folder for Docusaurus. Checkout how the other package readmes get released.

smorrisods (changes requested):
Amazing work! Here are some last mile observations.  We can discuss them if you like, @erincarrieresontario.

packages/ontario-design-system-component-html-generator/generated-samples-refactor-check-2/ontario-accordion.html
erincarrieresontario: Good idea 👍

packages/ontario-design-system-component-html-generator/README.md
erincarrieresontario: Okay great - done 👍

packages/ontario-design-system-component-html-generator/src/cli.ts
```diff
@@ -0,0 +1,163 @@
...
+import { generateSamples } from './index.js';
+import type { ComponentSample, SampleFormatterOptions } from './types.js';
+
+const INIT_TEMPLATE = `/**
+ * Ontario Design System — Sample Generator config
+ *
+ * Each entry in the \`samples\` array defines one HTML sample to generate.
+ * Run the generator with:
+ *
+ *   ontario-design-system-component-html-generator --samplesFile ./samples.config.js --outputDirectory ./output
```
smorrisods: This name is _really_ long for a CLI tool, perhaps we can shorten it to something like `odscompgen` or `odscompconvert` or something like that.

packages/ontario-design-system-component-html-generator/src/index.ts
```diff
@@ -0,0 +1,199 @@
...
+ *
+ * Responsibilities:
+ * - Render input sample HTML using a pluggable renderer.
+ * - Normalize/clean the rendered markup.
+ * - Resolve and transform component styles.
+ * - Return a deterministic result object (`items` + `summary`) without doing filesystem I/O.
+ *
+ * Filesystem writes and batch orchestration are intentionally handled by the CLI layer.
+ */
+
```
smorrisods: There appears to be an extra line here between the JSDoc comment and the...wait, what is this for?

What is this JSDoc referencing?

```suggestion
```

packages/ontario-design-system-component-html-generator/src/types.ts
```diff
@@ -0,0 +1,115 @@
+/**
+ * Public type contracts for `@ongov/app-sample-generator`.
+ *
+ * These interfaces define the library-first API:
+ * callers provide a configuration object and receive a deterministic result object,
+ * while transport concerns (filesystem, HTTP, etc.) are handled by consumers.
+ */
```
smorrisods: This is using JSDocs when it looks more like just regular comments? Is this a normal practice for file level documentation?

packages/ontario-design-system-component-html-generator/src/utils.test.ts
```diff
@@ -0,0 +1,241 @@
...
+describe('formatHtml', () => {
+	it('removes DOCTYPE, html, head, and body wrappers', () => {
+		const input = '<!doctype html><html><head><title>x</title></head><body><p>content</p></body></html>';
+		const result = formatHtml(input);
+		expect(result).not.toMatch(/<!doctype/i);
+		expect(result).not.toMatch(/<html/i);
+		expect(result).not.toMatch(/<head/i);
+		expect(result).not.toMatch(/<body/i);
+		expect(result).toContain('<p>');
+	});
```
smorrisods, line 195: I'm a little confused `formatHTML` is actively stripping the HTML, and `buildDocument` is adding it back if requested via `fullDocument`? Could the HTML not be stripped if this option is set?

smorrisods (changes requested):
I've added a few more comments.

packages/ontario-design-system-component-html-generator/src/utils.test.ts
erincarrieresontario: Yeah so we are stripping the renderer’s HTML wrapper on purpose, then we add back our own wrapper when `fullDocument` is enabled. That keeps the final output consistent and avoids depending on whatever boilerplate the renderer returned.

packages/ontario-design-system-component-html-generator/src/types.ts
erincarrieresontario: Yep good catch, switching to reg comments

packages/ontario-design-system-component-html-generator/src/cli.ts
erincarrieresontario: Totally - changed this to odscompgen!

packages/ontario-design-system-component-html-generator/src/utils.test.ts
smorrisods: That sounds sensible, it gives us the control to recompose things as we see fit.

packages/ontario-design-system-component-html-generator/src/index.ts
```diff
@@ -0,0 +1,198 @@
...
+ * Library entrypoint for sample generation.
+ *
+ * Responsibilities:
+ * - Render input sample HTML using a pluggable renderer.
+ * - Normalize/clean the rendered markup.
+ * - Resolve and transform component styles.
+ * - Return a deterministic result object (`items` + `summary`) without doing filesystem I/O.
+ *
+ * Filesystem writes and batch orchestration are intentionally handled by the CLI layer.
+ */
```
smorrisods: 🤔 If this is file level it can go above the imports, like is already done in `packages/ontario-design-system-component-html-generator/src/sample-config.ts` 🙂.

smorrisods (changes requested):
It's looking amazing! The only last thing I can think of is JSON support for the CLI so full TS files aren't required for processing, is this supported?

erincarrieresontario:
> It's looking amazing! The only last thing I can think of is JSON support for the CLI so full TS files aren't required for processing, is this supported?

JSON support is fully in! You can now pass a .json file directly to `--samplesFile,  it accepts either a bare array or an object with a `samples` key. I also added a `--format json` flag to the init command so you can scaffold a JSON starter file without needing to write any JS/TS at all

smorrisods (approved):
LGTM 🚀

PR #187, opened 2026-02-24
DS-2430: Add VRT and E2E tests for ontario-page-alert

MeaghanCarrieres:
This MR adds Next.js SSR support for the Ontario Design System React component library, resolves styling issues with slotted page alert content, and introduces VRT + E2E test coverage for the `ontario-page-alert` component.

### app-nextjs

- Configured Sass `pkg:` resolution using new exported `pkgImporter`
- Added local `ontario-theme.scss` wrapper:
   - forwards global theme with `$asset-base-path`
   - includes slotted styles for callouts/asides and page alerts
- Imported theme wrapper in `layout.tsx`
- Added VRT coverage for page alert variants
- Added E2E tests validating:
    - variant rendering
    - slot vs prop behaviour
    - edge cases (missing heading, invalid type fallback)
    - hydration behaviour in Next.js

### Component Library

- Updated `ontario-page-alert` "type" prop to include `reflect: true`
    - ensures attribute presence for CSS selectors targeting slotted content

### React Component Library

- Added reusable `pkgImporter` helper for Next.js Sass resolution
- Exported helper via package exports
- Updated README with Next.js installation and configuration guidance

## Why this change was needed

1. Next.js does not resolve Sass `pkg:` imports by default, causing theme and slotted styles to fail compilation.
2. Slotted page-alert styles rely on attribute selectors (`ontario-page-alert[type="error"]`), requiring the prop to be reflected to the DOM.
3. SSR support is required for correct hydration and declarative shadow DOM behaviour in Next.js.
4. Additional test coverage ensures page alert variants and slot styling regressions are caught.

## Result

- Next.js app builds and hydrates successfully
- Theme and slotted styles compile correctly
- Page alert styling behaves as expected across variants
- Consumers now have documented Next.js setup steps
- Page alert behaviour is protected by VRT and E2E coverage


packages/ontario-design-system-component-library-react/src/next/sass-pkg-importer.ts
smorrisods, line 1: Oh, you had to write your own?

packages/ontario-design-system-component-library-react/src/next/sass-pkg-importer.ts
```diff
@@ -0,0 +1,25 @@
...
+	} catch {}
+
+	try {
+		const parts = spec.split('/');
+		const base = parts.pop();
+		const dir = parts.join('/');
+		return { file: require.resolve(`${dir}/_${base}.scss`) };
+	} catch {}
+
+	return null;
```
smorrisods, line 70: All errors are swallowed here.  If nothing resolves it might be worth us knowing.  One idea is to add logging but take in an option `debug` parameter that can let us enable the logging if needed.

packages/ontario-design-system-component-library-react/README.md
```diff
@@ -32,6 +34,21 @@ To use the Ontario Design System React component library, follow these steps:
...
+
+   ```scss
+   // src/styles/ontario-theme.scss
+   @forward 'pkg:@ongov/ontario-design-system-global-styles/styles/scss/theme.scss' with (
+   	$asset-base-path: '/assets'
+   );
+
+   // Include slotted styles
+   @use 'pkg:@ongov/ontario-design-system-component-library-react/styles/slotted-styles/callouts-asides.scss';
+   @use 'pkg:@ongov/ontario-design-system-component-library-react/styles/slotted-styles/page-alerts.scss';
```
smorrisods: I wonder if it is worth putting into the React lib and then just documenting how to update the asset path with `with`.  What do you think?

smorrisods (changes requested):
Sorry, @MeaghanCarrieres apparently I didn't submit my comments 😅

packages/ontario-design-system-component-library-react/src/next/sass-pkg-importer.ts
MeaghanCarrieres: Yeah, I had to write one to get it to work for the Next app!

packages/ontario-design-system-component-library-react/README.md
MeaghanCarrieres: Do you mean having this file as something importable by the react library but adding a customization for the `asset-base-path` value?

packages/ontario-design-system-component-library-react/src/next/sass-pkg-importer.ts
MeaghanCarrieres: Updated!

packages/ontario-design-system-component-library-react/README.md
MeaghanCarrieres: Hey @smorrisods,

I looked into this a bit more and realized the React library is already exporting a `theme.scss` file that forwards the global theme and includes the slotted styles (it’s autogenerated during the build).

Because of that, consumers don’t actually need to recreate the wrapper we had in the README. They can just forward the exported theme and override `$asset-base-path` if needed.

I updated the README to show this simpler pattern:

```
@forward 'pkg:@ongov/ontario-design-system-component-library-react/styles/theme.scss' with (
  $asset-base-path: '/your-assets-path'
);
```

I figure this way:
- users only need to override the one variable they care about
- we don’t have to document all the slotted style imports in the README
- if additional slotted styles are added later, the autogenerated theme file will include them automatically.

What do you think?

packages/ontario-design-system-component-library/src/components/ontario-page-alert/ontario-page-alert.tsx
```diff
@@ -31,25 +31,26 @@ export class OntarioPageAlert implements PageAlert {
...
 	 */
 	@Prop() content: string;
 
+	/**
+	 * Internal, sanitized type used for rendering.
+	 * (So invalid values fall back to informational.)
+	 */
+	@State() private typeState: PageAlertType = 'informational';
+
+	private readonly validTypes: PageAlertType[] = ['informational', 'warning', 'success', 'error'];
```
smorrisods: There is a `ontario-page-alert.interface.ts` interface file with the type definition, we can model the `packages/ontario-design-system-component-library/src/components/ontario-button/ontario-button.types.ts` for creating a single type source of truth for `PageAlertType`.

smorrisods (changes requested):
Only one small comment before 🚀

packages/ontario-design-system-component-library/src/components/ontario-page-alert/ontario-page-alert.tsx
MeaghanCarrieres: Updated!

packages/ontario-design-system-component-library-react/README.md
smorrisods: This is a good solution, thanks for the investigation and finding out out a way to keep things succinct.

packages/ontario-design-system-component-library/src/components/ontario-page-alert/ontario-page-alert.tsx
smorrisods: Awesome!

smorrisods (approved):
LGTM! 🚀

PR #197, opened 2026-03-17
DS-2574: Swap Topics and Sign-in button order and improve overflow menu accessibility

MeaghanCarrieres:
This MR updates the header to swap the order of the **Topics** and **Sign-in** menu buttons.

During implementation, it became clear that the existing keyboard navigation and focus management logic in the header overflow menus did not support this change correctly. Additional updates were required to ensure accessibility and maintain expected keyboard behaviour across desktop and mobile/tablet views.

### Header updates
- Swapped the order of **Topics** and **Sign-in** menu buttons in the `ontario-header` component
- Updated related logic to ensure correct menu toggling and focus restoration

### Keyboard navigation improvements (desktop / standalone mode)
- Added support for returning focus to the trigger button when tabbing from the last menu item
- Introduced new prop on `ontario-header-overflow-menu`:
     - Added `returnFocusToTriggerOnLastTab` prop. When enabled, tab on the last menu item:
         - prevents default behaviour
         - emits `focusMenuButton` event
         - closes the menu
         - restores focus to the correct trigger (in this case, either **Topics** or **Sign-in**)
     - This ensures:
         - Focus does not move unexpectedly to the next header control
         - Each menu behaves independently and predictably

### Embedded mode fixes (mobile / tablet)
- Fixed issue where pressing ArrowDown / ArrowUp would skip the first menu item
    - Root cause: the same key event was being handled twice:
         - once to focus the first item (via parent tabs)
         - once again by the menu’s arrow navigation
    - Fix:
        - Added a suppression flag to ignore the initial arrow event after programmatic focus
    - Ensures:
        - First ArrowDown focuses the first item (not the second)
        - Arrow navigation behaves consistently in embedded mode

### Focus management improvements
- Maintained separation of concerns:
    - `ontario-header-overflow-menu` remains agnostic
    - `ontario-header` continues to own focus restoration logic
    - Reused existing `focusMenuButton` event for both:
        - Shift+Tab from first item
        - Tab from last item (new behaviour)


packages/ontario-design-system-component-library/src/components/ontario-header/ontario-header.tsx
```diff
@@ -595,9 +598,14 @@ export class OntarioHeader {
...
-			this.menuButton.focus();
-		}
+		// if (this.menuButton) {
+		// 	this.menuButton.focus();
+		// }
+		// if (this.signInButton) {
+		// 	this.signInButton.focus();
+		// } else {
+		// 	this.menuButton.focus();
+		// }
```
erincarrieresontario: looks like we are commenting out everything in this function, was this intended?

erincarrieresontario (approved):
Look good @MeaghanCarrieres just one comment!

packages/ontario-design-system-component-library/src/components/ontario-header/ontario-header.tsx
MeaghanCarrieres: Good catch!! I forgot to remove this.

PR #204, opened 2026-03-25
DS-2607: add aggregate value API to ontario-date-input

smorrisods:
## Summary

- add an aggregate `value` API to `ontario-date-input` and hydrate the internal fields from ISO input
- align form components around host `value`, host-native events, and richer event detail where it adds value
- extend the work to `textarea`, `dropdown-list`, `radio-buttons`, and `checkboxes` so the public value contract is more consistent
- add stronger unit coverage and modern Playwright e2e coverage for the updated controls

## Notes

- this PR keeps the existing component-specific custom events, but makes the host `value` and host event story more predictable
- legacy commented e2e tests were left in place under the new Playwright tests so the older cases can be ported over incrementally


packages/ontario-design-system-component-library/src/components/ontario-checkbox/test/ontario-checkboxes.e2e.ts
```diff
@@ -1,3 +1,66 @@
...
+		checkboxes = host.locator('input[type="checkbox"]');
+		labels = host.locator('.ontario-checkboxes__label');
+		changeSpy = await page.spyOnEvent('change');
+		checkboxOnChangeSpy = await page.spyOnEvent('checkboxOnChange');
+	});
+
+	test('updates the component value and emits host/custom change details when checkboxes are toggled', async ({
+		page,
+	}) => {
+		await labels.nth(0).click();
```
erincarrieresontario, line 33: Should we also check for unchecking?

packages/ontario-design-system-component-library/src/components/ontario-date-input/ontario-date-input.tsx
```diff
@@ -231,6 +248,45 @@ export class OntarioDateInput {
 		this.updateCaptionState(this.caption);
 	}
 
+	@Watch('value')
+	syncValueProp(newValue?: string) {
```
erincarrieresontario, line 252: Right now the watcher assumes value is always a string, but the prop actually accepts both string and string[]. If someone passes an array (which is valid), the watcher gets the array and things break.

erincarrieresontario (commented):
Looks good @smorrisods , just a few comments!

packages/ontario-design-system-component-library/src/components/ontario-checkbox/test/ontario-checkboxes.e2e.ts
smorrisods: Good catch! I'll update it in my next fly-by 🛰️

packages/ontario-design-system-component-library/src/components/ontario-date-input/ontario-date-input.tsx
smorrisods: Good catch! I'll review this and orbit back 🪐

smorrisods: Looking a the definition of `value` it only takes in a `string` so I'll guard against this case you've noticed.

smorrisods:
@erincarrieresontario I've made updates 🙂

packages/ontario-design-system-component-library/src/components/ontario-date-input/test/ontario-date-input.e2e.ts
```diff
@@ -1,3 +1,71 @@
...
+		fieldInputSpy = await page.spyOnEvent('inputOnInput');
+	});
+
+	test('normalizes the component value after a complete date is entered', async ({ page }) => {
+		const yearInput = host.locator('input').nth(0);
+		const monthInput = host.locator('input').nth(1);
+		const dayInput = host.locator('input').nth(2);
+
+		await yearInput.fill('2024');
+		await page.waitForChanges();
```
erincarrieresontario: Do we need all of these, waitForChanges? Stencil’s page.waitForChanges() is needed only when the component updates asynchronously, ,ost .fill() calls already trigger input events synchronously.

packages/ontario-design-system-component-library/src/components/ontario-date-input/test/ontario-date-input.spec.tsx
```diff
@@ -108,14 +108,255 @@ describe('ontario-date-input', () => {
...
+		const page = await newSpecPage({
+			components: [OntarioDateInput],
+			html: `<ontario-date-input></ontario-date-input>`,
+		});
+
+		const inputs = page.root?.shadowRoot?.querySelectorAll('input');
+		const [yearInput, monthInput, dayInput] = Array.from(inputs ?? []);
+
+		yearInput.value = '2024';
+		yearInput.dispatchEvent(new Event('input'));
```
erincarrieresontario: These tests are dispatching plain Event('input'), which doesn’t behave like a real browser input event. Real input events bubble and are composed, and Stencil relies on that. We should switch to new InputEvent('input', { bubbles: true, composed: true }) so the tests accurately simulate user interaction.

packages/ontario-design-system-component-library/src/components/ontario-radio-buttons/test/ontario-radio-button.e2e.ts
```diff
@@ -1,3 +1,60 @@
...
+					{ "value": "radio-option-2", "elementId": "radio-2", "label": "Radio option 2 label" }
+				]'
+			></ontario-radio-buttons>
+		`);
+		await page.waitForChanges();
+
+		host = page.locator('ontario-radio-buttons').first();
+		radios = host.locator('input[type="radio"]');
+		labels = host.locator('.ontario-radios__label');
+		changeSpy = await page.spyOnEvent('change');
```
erincarrieresontario, line 26: minor thing, spying on the native change event at the page level can pick up unrelated events. It’s safer to scope the spy to the host element (host.spyOnEvent(...)) so we’re only asserting events emitted by this component.

smorrisods: We're using the `page` to spy because Stencil's tooling is provided on `page`.  The assertions should be tight enough that other noise from other isn't an issue.

erincarrieresontario (approved):

PR #205, opened 2026-03-25
DS-2608: clarify step indicator precedence and warnings

smorrisods:
## Summary

- make `percentageComplete` take explicit precedence over step props, including `0`
- warn when step and percentage props are mixed, when step props are incomplete, and when back-button props are ambiguous or non-actionable
- keep the requested back-control render path visible while surfacing misconfiguration through the design system warning pattern
- update the step indicator README and spec coverage to document and verify the new precedence rules

## Notes

- `percentageComplete` now uses defined-value precedence instead of truthiness, so `0` renders correctly in percentage mode
- incomplete step-mode props now render a `?` placeholder and emit a warning
- back controls still render when requested, but ambiguous or non-actionable combinations now emit warnings


packages/ontario-design-system-component-library/src/components/ontario-step-indicator/test/ontario-step-indicator.spec.tsx
```diff
@@ -0,0 +1,173 @@
...
+		page.rootInstance.customOnClick = jest.fn();
+		warnSpy.mockClear();
+		page.rootInstance.componentWillLoad();
+		await page.waitForChanges();
+
+		expect(page.root?.shadowRoot?.querySelector('a')).not.toBeNull();
+		expect(page.root?.shadowRoot?.querySelector('button')).toBeNull();
+		expect(warnSpy).toHaveBeenCalledTimes(1);
+	});
+});
```
erincarrieresontario, line 227: Maybe we could group some of the related tests with some 'desbribe' blocks? just to help improve readability.

erincarrieresontario (approved):
LGTM just one suggestion!

PR #228, opened 2026-05-07
DS-2246: Mark decorative icons as non-focusable for accessibility

muntalee:
Decorative SVG icons used in components like ontario-hint-expander should not be keyboard-focusable to comply with accessibility standards. These icons are already marked with aria-hidden="true" to hide them from screen readers.

- Update icon generator template to inject `focusable="false"` on all decorative icons
- Update chevron-up and chevron-down icons to include `focusable="false"` attribute
- Update unit tests and snapshots to reflect the new attribute

JIRA: DS-2246


smorrisods:
Thanks for your first PR, @muntalee!

Erin and I will review it and give you any feedback here on GitHub.  One stylistic note, we generally don't use Conventional Commit prefixes for our PR titles.  If there is a Jira issue then it should prefix the title.  Also, the title should be descriptive of what the PR is bringing to the code base, as well as start with a capital letter.

Eg. ~~fix(ontario-icon): mark decorative icons as non-focusable for accessibility~~ → DS-2246: Mark decorative icons as non-focusable for accessibility

For each PR, try and label it with the appropriate labels.  At a minimum, label a PR with what package is touched by the code in the PR. Eg. this PR touches the Component Library package so it would get the _component library_ label.

erincarrieresontario:
Hey @muntalee , Great work :) Just some suggestions for this ticket:
- I think that we may actually want to add the `isDecorative` check to the `focusable` attribute here (since we don't want to always make this icon not focusable, only when it is decorative or within the hint-text component). Then, we would actually want to pass this prop in within the ontario-hint-expander component, something like this: 
`<ontario-icon-chevron-down colour="inherit" is-decorative="true"></ontario-icon-chevron-down>`
- Also, for the branch name, it shoud look something like this: `munta/bug/DS-2246-accessibility-updates-to-chevron-icons`

erincarrieresontario (changes requested):
See feedback provided in comments

smorrisods:
>     * Also, for the branch name, it shoud look something like this: `munta/bug/DS-2246-accessibility-updates-to-chevron-icons`

Thanks for catching this, @erincarrieresontario.  In the case of the currently opened PRs changing the branch name will require new PRs to be opened.  For now, @muntalee, let's keep them and add the description going forward with new branches.

muntalee:
Apologies for the branch naming, I'll keep that in mind for future PRs. I sent a commit with the changes Erin suggested.

erincarrieresontario (approved):
LGMT 👍 Thanks for making those changes! I'll let @smorrisods also have a look!

PR #230, opened 2026-05-08
DS-2625: Prevent active background gaps on uneven card heights

muntalee:
- ensure the white text surface stretches to the bottom of basic cards
- prevent active grey from showing through when card content heights differ
- scope flex-fill behavior to basic cards to avoid horizontal layout regressions

JIRA: DS-2625

### DIFF
--- a/packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.scss
+++ b/packages/ontario-design-system-component-library/src/components/ontario-card/ontario-card.scss
@@ -14,6 +14,7 @@
 	border-radius: globalVariables.$global-radius;
 	margin-bottom: spacing.$spacing-7;
 	padding: spacing.$spacing-0;
+	overflow: hidden;
 	/*
 	 * Required for cards to stay the same height regardless of content size.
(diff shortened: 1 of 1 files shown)


smorrisods (approved):
LGTM, approved.  I'll let @erincarrieresontario take a look as well before merging.

erincarrieresontario (commented):
Can we actually just have one style definition for the `.ontario-card__card-type--basic` where it is : 
```
.ontario-card__card-type--basic {
	display: flex;
        flex: 1 1 auto; 
	flex-direction: column;
}
```

Or do we need the multiple nested definitions?

muntalee:
> Can we actually just have one style definition for the `.ontario-card__card-type--basic` where it is :
> 
> ```
> .ontario-card__card-type--basic {
> 	display: flex;
>         flex: 1 1 auto; 
> 	flex-direction: column;
> }
> ```
> 
> Or do we need the multiple nested definitions?

I kept the nested definitions because the base `.ontario-card__card-type--basic` only sets the layout direction. The white background reaching the bottom depends on the child elements growing to fill the remaining space, so collapsing it into one rule would risk breaking that behavior.

PR #232, opened 2026-05-11
DS-2634: Remove duplicate focus stop on image cards

muntalee:
- Removed the image anchor wrapper from ontario-card so image cards no longer create a second tab stop.
- Kept the heading link as the single keyboard-focus target.
- Preserved click-through behavior for the card.
- Added unit coverage for single-link rendering and href preservation.
- Updated the card accessibility guidance in the README.


erincarrieresontario (changes requested):
Hi @muntalee , I don't think we want to remove the anchor tag from the image, as we still want the image to be clickable. I think we can remove the focus line with CSS :)

erincarrieresontario (approved):
LGTM 👍 Will also let @smorrisods review!

smorrisods (approved):
LGTM 🚀

PR #233, opened 2026-05-13
Fix incorrect setAssetPath value and asset copy paths in React and Angular READMEs (DS-2641)

smorrisods:
## Summary

- Corrects the documented `setAssetPath` value from `${window.location.origin}/assets/` to `${window.location.origin}` in both the React and Angular README files
- Fixes the `copy:images` glob paths, which referenced a non-existent `dist/component-library/assets/` path
- Fixes the one-liner copy command to use `dist/assets/images/**` with `-f` so images land flat in the destination
- Updates the React copy destination from `src/assets` to `public/assets` to match Vite/CRA convention

## Why

The `/assets/` suffix in `setAssetPath` caused Stencil's URL resolution to produce a double-segment path (`/assets/assets/foo.svg`), resulting in a silent 404 for all component images. This was the root cause of the missing dropdown arrow reported in DS-2641.

Our internal sample apps worked because they already used `${window.location.origin}` (no suffix) — the docs were simply wrong.

## Details

Stencil's `getAssetPath('./assets/foo.svg')` appends relative to the `setAssetPath` base using standard URL resolution. With a trailing-slash base of `http://example.com/assets/`, `./assets/foo.svg` resolves to `/assets/assets/foo.svg`. Passing `window.location.origin` only gives the correct `/assets/foo.svg`.

This is a **documentation-only change** — no library code was modified.

## Validation

- Verified fix in consumer app (Avi's minimal test project): clean build, correct asset resolution
- Internal sample apps (React, Angular, Next.js) unaffected — they already used the correct pattern
- Docusaurus README copies will update automatically on next build

## Risks

None — docs only.

### DIFF
--- a/packages/ontario-design-system-component-library-angular/README.md
+++ b/packages/ontario-design-system-component-library-angular/README.md
@@ -96,10 +96,10 @@ If your app serves component assets (fonts, images, favicons) from a non-root pa
 ```typescript
 import { setAssetPath } from '@ongov/ontario-design-system-component-library-angular';
 
-setAssetPath(`${window.location.origin}/assets/`);
+setAssetPath(`${window.location.origin}`);
 ```
 
-## Usage
+If your app serves component assets (fonts, images, favicons) from a non-root path, configure the asset base path before components render. Pass `window.location.origin` (no `/assets/` suffix) — the library's internal asset helper appends the correct path segment automatically.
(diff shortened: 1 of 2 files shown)


erincarrieresontario (approved):
LGTM 👍

PR #242, opened 2026-05-21
DS-1931: Allow HTML/SVG content inside ontario-button via slot

muntalee:
Fixes HTML/SVG content being stripped when passed as children to `ontario-button`. The component read `host.textContent` and rendered it as a plain string, discarding any markup.

- Render `<slot>` instead of the text string when no `label` prop is set
- Suppress the missing-label warning when `aria-label-text` is provided
- Vertically align slotted `<svg>` and `` elements

## Example:
```html
<ontario-button aria-label-text="Open map">
  <svg aria-hidden="true" focusable="false">...</svg>
</ontario-button>
```

### DIFF
--- a/packages/ontario-design-system-component-library/src/components/ontario-button/ontario-button.scss
+++ b/packages/ontario-design-system-component-library/src/components/ontario-button/ontario-button.scss
@@ -51,6 +51,11 @@ $ontario-button-bg-secondary: colours.$ontario-colour-white;
 	.ontario-icon {
 		margin-right: 4px;
 	}
+
+	::slotted(svg),
+	::slotted(img) {
+		vertical-align: middle;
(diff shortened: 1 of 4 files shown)


smorrisods:
This change is a fundamental structural change to our button, one which needs discussion with the greater team.

Good work here, @muntalee! We can't proceed with it without higher direction.

smorrisods:
Idea:

- Place the icon in a slot and have a prop to decide where to place the slot, left or right.
- Add two CSS custom vars to override the padding on either side of the icon within the button (icons have a fixed square bounding box size)
- Profit‽?

smorrisods:
@muntalee I have some updates for you we should chat about when you get some time!

PR #243, opened 2026-05-22
DS-2236: Fix highlight prop breaking `ontario-table` in Next.js SSR

muntalee:
Fixes `ontario-table` breaking in Next.js SSR environments when `highlight: true` is set on a `tableData` row.

- Replace `getAssetPath` highlight indicator image with an inline SVG
- Move SVG fill colour to `colours.$ontario-colour-yellow` design token in SCSS
- Initialize state arrays to `[]` to prevent undefined errors during SSR hydration
- Add null guard to `applyScrollbar` to prevent errors before DOM refs are assigned
- Update snapshot to reflect inline SVG output

### DIFF
--- a/packages/ontario-design-system-component-library/src/components/ontario-table/test/__snapshots__/ontario-table.spec.tsx.snap
+++ b/packages/ontario-design-system-component-library/src/components/ontario-table/test/__snapshots__/ontario-table.spec.tsx.snap
@@ -226,7 +226,7 @@ exports[`ontario-table snapshot - 5 columns (numeric) should render the expected
               <th scope="row">
                 Forestry
                 <svg aria-hidden="true" class="ontario-table--highlight-indicator" focusable="false" viewBox="0 0 11.3 22.6" xmlns="http://www.w3.org/2000/svg">
-                  <polygon fill="#FCAF17" points="0,0 11.3,11.3 0,22.6"></polygon>
+                  <polygon points="0,0 11.3,11.3 0,22.6"></polygon>
                 </svg>
               </th>
               <td class="ontario-table-cell--numeric">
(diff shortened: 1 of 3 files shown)


packages/ontario-design-system-component-library/src/components/ontario-table/test/__snapshots__/ontario-table.spec.tsx.snap
```diff
@@ -225,7 +225,9 @@ exports[`ontario-table snapshot - 5 columns (numeric) should render the expected
             <tr class="ontario-table-row--highlight">
               <th scope="row">
                 Forestry
-                <img aria-hidden="true" class="ontario-table--highlight-indicator" src="/assets/highlight-indicator.svg">
+                <svg aria-hidden="true" class="ontario-table--highlight-indicator" focusable="false" viewBox="0 0 11.3 22.6" xmlns="http://www.w3.org/2000/svg">
+                  <polygon points="0,0 11.3,11.3 0,22.6"></polygon>
+                </svg>
               </th>
```
smorrisods: Let's switch this to use the `getImageAssetSrcPath` helper function from 'utils/helper/assets`.  This keeps the component design in line with our Header, Footer, and Dropdown List components that also use SVG assets.

The design here is good and is worth considering in a future update, but for now let's follow our existing pattern.

smorrisods (changes requested):

packages/ontario-design-system-component-library/src/components/ontario-table/test/__snapshots__/ontario-table.spec.tsx.snap
muntalee: Thank you for that, didn't notice the script. Added the change in 5fc2597.

smorrisods (approved):
LGTM 🚀

PR #245, opened 2026-05-25
Fix `app-angular` compile error after `@ngx-translate/http-loader` v17 upgrade

muntalee:
The Angular POC fails to compile due to a breaking API change in
`@ngx-translate/http-loader` v17. The `TranslateHttpLoader` constructor
no longer accepts arguments — configuration is now injected via
`provideTranslateHttpLoader()`.

This was caused by Renovate bumped the package to v17 in `e2861e4b` but the call site in
`translation.config.ts` was not updated at the time.

- Removed the `createTranslateLoader` factory function from `translation.config.ts`
- Updated `app.module.ts` to use `useClass: TranslateHttpLoader` and
  `provideTranslateHttpLoader()` with the existing translation path config

## Testing

Run `pnpm run serve` from `packages/app-angular` and confirm the app compiles
and loads at http://localhost:4200.

### DIFF
--- a/packages/app-angular/src/app/app.module.ts
+++ b/packages/app-angular/src/app/app.module.ts
@@ -23,9 +23,11 @@ import { FrameSixComponent } from './pages/framesix/framesix.component';
 
 // import ngx-translate and the http loader
 import { TranslateLoader, TranslateModule } from '@ngx-translate/core';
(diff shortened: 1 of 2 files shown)


erincarrieresontario (approved):
LGTM 👍

smorrisods (approved):
LGTM 🚀

PR #248, opened 2026-05-27
DS-1762: Fix htmlType submit and reset for OntarioButton

muntalee:
OntarioButton has a `htmlType` prop for form buttons, but it did not actually work inside a form because the component renders in Shadow DOM. This change fixes that by making submit and reset work the way users expect.

- Use `form.requestSubmit()` when `htmlType="submit"`
- Use `form.reset()` when `htmlType="reset"`
- Check `htmlType` at click time so later prop changes still work

- A form demo and Playwright tests were added so we can verify both cases:
    - Submit shows that the form handler runs
    - Reset shows that the input is cleared

### DIFF
--- a/packages/app-nextjs/src/app/components/ontario-button/page.tsx
+++ b/packages/app-nextjs/src/app/components/ontario-button/page.tsx
@@ -1,9 +1,12 @@
 'use client';
 
+import { useState } from 'react';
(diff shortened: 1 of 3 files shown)


smorrisods (approved):
Good catch and update! 🚀

PR #250, opened 2026-05-27
DS-1760: Make OntarioInput `defaultValue` work for React uncontrolled forms

muntalee:
Fixes OntarioInput so React users can pass `defaultValue` and use the component in uncontrolled forms as expected. It also keeps controlled behavior intact, where value still takes precedence when both props are provided.

- Added `defaultValue` support to OntarioInput and mapped it to the internal input behavior.
- Initialized component value from `defaultValue` when value is not provided.
- Kept controlled precedence so value overrides `defaultValue` when both are set.
- Added unit tests for `defaultValue` initialization and value/defaultValue precedence.
- Updated generated wrappers, typings, and docs so `defaultValue` is exposed consistently across packages.


erincarrieresontario (approved):
👍

smorrisods:
Closing due to performing a greater review of React integrations.

PR #251, opened 2026-05-28
DS-1761: Fix OntarioInput React ref mapping

muntalee:
Fixes OntarioInput ref behavior for React so refs resolve to the native input element instead of the web component host, enabling uncontrolled form usage and smoother third-party form library integration.

- Added `getInputElement()` as a public method on `ontario-input` to expose the internal native input
- Stored a stable internal input reference in the component and returned it from `getInputElement()`
- Added unit tests for method identity and regression coverage for normal input interactions
- Added a focused React `OntarioInput` override that forwards refs to the native input via `getInputElement()`
- Kept generated React files untouched by implementing the override in non-generated source and export wiring

### DIFF
--- a/packages/ontario-design-system-component-library-react/src/overrides/ontario-input-ref.tsx
+++ b/packages/ontario-design-system-component-library-react/src/overrides/ontario-input-ref.tsx
@@ -32,7 +32,7 @@ const mapHostRefToNativeInput = (hostElement: HTMLElement | null, ref: React.For
 	hostWithInputMethod
 		.componentOnReady?.()
 		.then(() => assignInputRef())
-		.catch((_error: unknown): void => undefined);
+		.catch(() => {});
 };
 
 type OntarioInputProps = React.ComponentPropsWithoutRef<typeof GeneratedOntarioInput>;
--- a/packages/ontario-design-system-component-library-react/src/overrides/ontario-input-ref.tsx
+++ b/packages/ontario-design-system-component-library-react/src/overrides/ontario-input-ref.tsx
@@ -1,3 +1,5 @@
+'use client';
+
 import React from 'react';
 
 import { OntarioInput as GeneratedOntarioInput } from '../components/components.js';
--- a/packages/ontario-design-system-component-library-react/src/overrides/ontario-input-ref.tsx
+++ b/packages/ontario-design-system-component-library-react/src/overrides/ontario-input-ref.tsx
@@ -1,0 +1,54 @@
+import React from 'react';
+
+import { OntarioInput as GeneratedOntarioInput } from '../components/components.js';
+import { setRef } from '../react-component-lib/utils/index.js';
+
+type OntarioInputHostElement = HTMLElement & {
+	getInputElement?: () => Promise<HTMLInputElement | null>;
+	componentOnReady?: () => Promise<unknown>;
+};
+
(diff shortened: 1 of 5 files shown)


packages/ontario-design-system-component-library-react/src/overrides/ontario-input-ref.tsx
```diff
@@ -0,0 +1,56 @@
...
+		const inputElement = await hostWithInputMethod.getInputElement?.();
+		setRef(ref, inputElement ?? null);
+	};
+
+	assignInputRef().catch((_error: unknown): void => undefined);
+
+	hostWithInputMethod
+		.componentOnReady?.()
+		.then(() => assignInputRef())
+		.catch((_error: unknown): void => undefined);
```
erincarrieresontario:
```suggestion		
        .catch(() => {});
```

packages/ontario-design-system-component-library/src/components/ontario-input/test/ontario-input.spec.tsx
```diff
@@ -124,5 +124,44 @@ describe('ontario-input', () => {
...
+
+		it('should return the internal input element when using getInputElement', async () => {
+			const page = await newSpecPage({
+				components: [OntarioInput],
+				html: `<ontario-input
+					name="input-name"
+					element-id="input-id"
+				></ontario-input>`,
+			});
+
```
erincarrieresontario, line 136: Maybe add in a: 

```suggestion

expect(page.root).not.toBeNull();

```

packages/ontario-design-system-component-library/src/components/ontario-input/test/ontario-input.spec.tsx
```diff
@@ -124,5 +124,44 @@ describe('ontario-input', () => {
...
+			shadowInput.dispatchEvent(new Event('input'));
+			await page.waitForChanges();
+
+			const inputElement = await page.root?.getInputElement();
+
+			expect(inputElement).toBe(shadowInput);
+			expect(inputElement?.value).toBe('updated by user');
+			expect(page.rootInstance.value).toBe('updated by user');
+		});
 	});
```
erincarrieresontario, line 182: We could maybe add a test case for if the input isnt ready yet

erincarrieresontario (changes requested):

packages/ontario-design-system-component-library-react/src/overrides/ontario-input-ref.tsx
muntalee: Changed that in a669bfe7

packages/ontario-design-system-component-library/src/components/ontario-input/test/ontario-input.spec.tsx
muntalee: Added in 69de7852

packages/ontario-design-system-component-library/src/components/ontario-input/test/ontario-input.spec.tsx
muntalee: Great idea! Added a regression test for the no-internal-ref path so getInputElement resolves to null when the internal input ref is unavailable. Included in commit f166d601.

smorrisods:
Closing due to performing a greater review of React integrations alongside #250

PR #256, opened 2026-06-05
DS-2663: Implement `ontario-in-page-navigation` host component

muntalee:
Implements `ontario-in-page-navigation` as a Stencil web component using slotted composition with `ontario-in-page-navigation-item` children.

- Host component shell with semantic nav structure, heading rendering, and list container
- `heading`, `headingLevel`, `noTopBorder`, `smoothScroll`, `skipLinkTarget`, and `language` props
- Smooth-scroll handling for in-page anchors with hash update behavior
- Skip link support to configurable main-content target
- Language support via app language handling and localized defaults
- Console warnings/validation for invalid `heading` and `headingLevel` inputs
- `inPageNavigation` i18n keys added to `global.i18n.json`
- Readme with usage and accessibility guidance
- Updated demo usage in `index.html`
- Minimal unit and E2E test coverage for host behavior
- Regenerated React and Angular output targets

## Out of scope
- `ontario-in-page-navigation-item` child component implementation details (DS-2665)
- React and Angular hand-authored wrappers


packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.scss
```diff
@@ -0,0 +1,61 @@
...
+.ontario-page-navigation {
+	border-top: globalFunctions.px-to-rem(4) solid colours.$ontario-colour-black;
+	border-bottom: globalFunctions.px-to-rem(4) solid colours.$ontario-greyscale-20;
+	padding: spacing.$spacing-0;
+	margin: spacing.$spacing-7 spacing.$spacing-0;
+}
+
+.ontario-page-navigation-content {
+	padding: spacing.$spacing-7 spacing.$spacing-0;
+	max-width: 46rem;
```
erincarrieresontario: This seems like a very big number, what is it referencing? Can we add a comment here.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
```diff
@@ -0,0 +1,253 @@
...
+				.addRegularText('are supported. The default value')
+				.addMonospaceText(' h2 ')
+				.addRegularText('is assumed.')
+				.printMessage();
+
+			this.headingLevel = 'h2';
+		}
+	}
+
+	private hasSlotContent(slotElement?: HTMLSlotElement): boolean {
```
erincarrieresontario, line 131: Lets add some comments for all of these private functions

erincarrieresontario (commented):
This looks really good @muntalee , just a few comments

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.scss
muntalee: So the 46rem value comes from the original Fractal in-page navigation styles, where it’s used as the max width for the content column. I couldn’t find a shared design token for that exact value, so I kept it to preserve the existing layout. I added a small comment here in 2ba0cfd, but this could be worth discussing in the future as to whether keeping it as is or worry more about it once we implement the design tokens.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
muntalee: Lovely idea, added some in 804a903.

erincarrieresontario (approved):
LGTM 👍  Let's get a review from @smorrisods as well :)

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/readme.md
```diff
@@ -0,0 +1,120 @@
...
+		{label: 'Angular', value: 'angular'},
+	]}
+	groupId="framework"
+	queryString="framework">
+<TabItem value="html">
+```
+
+```html
+<ontario-in-page-navigation heading="On this page">
+	<ontario-in-page-navigation-item label="Overview" href="#overview"></ontario-in-page-navigation-item>
```
smorrisods: This readme (and index.html) references ontario-in-page-navigation-item everywhere, but that component doesn't exist yet in this branch, it only lives on the stacked PR #257. As it stands this PR isn't really usable on its own since the docs and demo page point to an element that isn't registered. Are we planning to land these together, or should we hold off updating the examples until #257 merges?

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
```diff
@@ -0,0 +1,274 @@
...
+					}}
+				>
+					<div class="ontario-page-navigation-content" onClick={this.handleLinkClick}>
+						<slot name="heading">{this.renderHeading(headingText)}</slot>
+						<slot name="skip-link">
+							<a class="ontario-page-navigation__skip-link ontario-show-on-focus" href={skipLinkHref}>
+								{this.translations.inPageNavigation.skip[resolvedLanguage]}
+							</a>
+						</slot>
+						<ol class="ontario-page-navigation-list" role="navigation">
```
smorrisods: We've got role="navigation" on this ol, but it's already nested inside the nav with aria-label on line 252. That gives us two navigation landmarks in the same area, one labelled and one not, which could be confusing for screen reader users. Could we drop the role here since a plain list inside the existing nav landmark should be enough?

smorrisods (changes requested):
Nice work on this one, the host component structure and heading handling look solid. A couple of things I want to chat through before this lands, mainly around the item component dependency and the generated types.

One more thing that doesn't have a diff to comment on directly: components.d.ts wasn't regenerated for this new component. The React and Angular wrapper files reference Components.OntarioInPageNavigation, but that type doesn't exist in components.d.ts, which will break type checking for consumers. Can we run the stencil build/docs step again before merging so this file picks it up?

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/readme.md
muntalee: Great point, agreed. I updated this PR to remove the dependency on the stacked item component in examples. The readme and demo index now use slotted list/link examples so this PR is usable on its own without requiring #257 to land first.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
muntalee: I removed `role="navigation"` from the inner ordered list so there is only a single navigation landmark at the parent nav level in b91b031, which avoids duplicate landmark announcements for screen reader users.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
```diff
@@ -0,0 +1,274 @@
...
+import translations from '../../translations/global.i18n.json';
+
+export type OntarioInPageNavigationHeadingLevel = 'h2' | 'h3' | 'h4';
+
+@Component({
+	tag: 'ontario-in-page-navigation',
+	styleUrl: 'ontario-in-page-navigation.scss',
+	shadow: true,
+})
+export class OntarioInPageNavigation {
```
smorrisods, line 26: This component's class is missing a JSDoc description, which we're trying to introduce across all components to better improve the documentation. Checkout the descriptions from the other props/methods in this file for more detail.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/readme.md
```diff
@@ -0,0 +1,126 @@
...
+	<li><a href="#eligibility">Eligibility</a></li>
+</ontario-in-page-navigation>
+```
+
+<!-- Auto Generated Below -->
+
+## Properties
+
+| Property         | Attribute          | Description                                                           | Type                        | Default          |
+| ---------------- | ------------------ | --------------------------------------------------------------------- | --------------------------- | ---------------- |
```
smorrisods, line 129: There don't appear to be any properties that are required, what does the component do when the user doesn't give it anything? Does it gracefully render nothing?

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.scss
```diff
@@ -0,0 +1,62 @@
...
+	border-top: globalFunctions.px-to-rem(4) solid colours.$ontario-colour-black;
+	border-bottom: globalFunctions.px-to-rem(4) solid colours.$ontario-greyscale-20;
+	padding: spacing.$spacing-0;
+	margin: spacing.$spacing-7 spacing.$spacing-0;
+}
+
+.ontario-page-navigation-content {
+	padding: spacing.$spacing-7 spacing.$spacing-0;
+	// 46rem keeps the content column at the same reading width used across Ontario pages.
+	max-width: 46rem;
```
smorrisods: Do we have a constant for this in other components? This seems like a magic number.  Perhaps we should give it a name here in this file if this is the only use of it so it comes with context.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.scss
```diff
@@ -0,0 +1,62 @@
...
+@use 'pkg:@ongov/ontario-design-system-global-styles/styles/scss/1-variables/colours.variables' as colours;
+@use 'pkg:@ongov/ontario-design-system-global-styles/styles/scss/1-variables/typography.variables';
+
+:host {
+	display: block;
+}
+
+.ontario-page-navigation {
+	border-top: globalFunctions.px-to-rem(4) solid colours.$ontario-colour-black;
+	border-bottom: globalFunctions.px-to-rem(4) solid colours.$ontario-greyscale-20;
```
smorrisods: Same as the note on #257 for the item component's font-size/line-height - these carry meaning and repeat within this file (border width used for both border-top and border-bottom), so let's turn this into a named variable rather than a raw `px-to-rem(4)` magic number.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.scss
```diff
@@ -0,0 +1,62 @@
...
+		overflow: visible;
+		margin-top: spacing.$spacing-2;
+		padding: spacing.$spacing-1 spacing.$spacing-2;
+		background-color: colours.$ontario-colour-white;
+	}
+}
+
+.ontario-page-navigation-list {
+	padding: spacing.$spacing-0;
+	padding-left: spacing.$spacing-4 + globalFunctions.px-to-rem(4);
```
smorrisods: Same question as on #257's item.scss `padding-bottom`: this uses the exact same `spacing.$spacing-4 + globalFunctions.px-to-rem(4)` expression - do we have a spacing var that's already `4px` we could use here instead of `px-to-rem`?

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
```diff
@@ -0,0 +1,274 @@
...
+	@Prop() smoothScroll?: boolean = true;
+
+	/**
+	 * Language used for translated defaults.
+	 */
+	@Prop({ mutable: true }) language?: Language;
+
+	@State() translations: any = translations;
+
+	@State() private hasDefaultSlotContent = false;
```
smorrisods, line 67: The `@State` props are without JSDocs, these should be added as well.

Also the type of `translations` is `any`.  Let's try to avoid `any` and give it a type.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
```diff
@@ -0,0 +1,274 @@
...
+	/**
+	 * Handles an update to the language should the user request a language update from the language toggle.
+	 */
+	@Listen('headerLanguageToggled', { target: 'window' })
+	handleHeaderLanguageToggled(event: CustomEvent<HeaderLanguageToggleEventDetails>) {
+		this.language = validateLanguage(event.detail.newLanguage);
+	}
+
+	@Watch('heading')
+	validateHeading() {
```
smorrisods, line 89: This should have a JSDoc comment.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
```diff
@@ -0,0 +1,274 @@
...
+				.addMonospaceText(' heading ')
+				.addRegularText('for')
+				.addMonospaceText(' <ontario-in-page-navigation> ')
+				.addRegularText('was not provided. A localized default heading will be used.')
+				.printMessage();
+		}
+	}
+
+	@Watch('headingLevel')
+	validateHeadingLevel(newValue: string) {
```
smorrisods, line 106: This one too - in fact all the private methods below already have one, let's keep it consistent.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+		const nodes = slotElement.assignedNodes({ flatten: true });
+
+		return nodes.some((node) => {
+			if (node.nodeType === Node.ELEMENT_NODE) {
+				return true;
+			}
+
+			if (node.nodeType === Node.TEXT_NODE) {
+				return Boolean(node.textContent?.trim());
```
smorrisods: I'm curious the choice for using `Boolean` directly rather than relying on `.trim()` to return `""` (empty string) and directly looking for it or using an implicit truthy check?

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+		const targetElement = this.getAnchorTarget(href);
+		if (!targetElement) {
+			return;
+		}
+
+		event.preventDefault();
+		targetElement.scrollIntoView({ behavior: 'smooth' });
+		history.pushState(null, '', href);
+	};
```
smorrisods, line 227: Clever!

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/readme.md
```diff
@@ -0,0 +1,126 @@
...
+<TabItem value="html">
+```
+
+```html
+<ontario-in-page-navigation heading="On this page">
+	<li><a href="#overview">Overview</a></li>
+	<li><a href="#eligibility">Eligibility</a></li>
+	<li><a href="#apply">Apply</a></li>
+</ontario-in-page-navigation>
+```
```
smorrisods, line 45: When the nav items are merged in are we going to add both examples, with regular items and list item examples, or is the cannonical way we're going to demo this is with the nav items?

smorrisods (changes requested):
Good work! Here are some comments from my top-to-bottom review.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
muntalee: Thanks for flagging this! I added the component class JSDoc in a68b6230.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/readme.md
muntalee: Clarified the no-items behaviour and empty-state guidance in 43920d91.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.scss
muntalee: Agreed. I replaced the raw value with a named SCSS variable for context in 74bc9992.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.scss
muntalee: Updated this to a named border-width variable in 74bc9992.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.scss
muntalee: I refactored this to use spacing tokens and named variables in 74bc9992.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
muntalee: Done in a68b6230

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
muntalee: Done in a68b6230

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
muntalee: Done in a68b6230

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
muntalee: Updated to a direct trimmed string check in a68b6230.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
muntalee: Thanks! My butler assisted heavily.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/readme.md
muntalee: Added clarification on canonical authoring/examples in 43920d91.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
muntalee: I fix missing class JSDoc description in a68b6230 so that the component has a top-level API description consistent with other components.

erincarrieresontario:
@muntalee just a heads up there are some merge conflicts here

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+import { Language } from '../../utils/common/language-types';
+import { HeaderLanguageToggleEventDetails } from '../../utils/events/common-events.interface';
+import { validateLanguage, validateValueAgainstArray } from '../../utils/validation/validation-functions';
+import { ConsoleMessageClass } from '../../utils/console-message/console-message';
+import translations from '../../translations/global.i18n.json';
+
+export type OntarioInPageNavigationHeadingLevel = 'h2' | 'h3' | 'h4';
+
+@Component({
```
smorrisods, line 21: The class JSDoc is sitting between `@Component` and `export class OntarioInPageNavigation`, but every other component in this library puts the doc comment *above* `@Component` (see `ontario-accordion.tsx`, `ontario-header.tsx`, `ontario-badge.tsx`, etc.). It's also missing the "For component guidance, see:" links to the design system docs that the others include. Let's align both the placement and the format:

```suggestion
/**
 * Ontario In-Page Navigation presents a heading and ordered list of links that help users navigate long single-page content.
 *
 * For component guidance, see:
 * - https://designsystem.ontario.ca/components/detail/in-page-navigation.html
 * - https://designsystem.ontario.ca/developer-docs/components/ontario-in-page-navigation/
 */
@Component({
	tag: 'ontario-in-page-navigation',
	styleUrl: 'ontario-in-page-navigation.scss',
	shadow: true,
})
export class OntarioInPageNavigation {
```

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
```diff
@@ -0,0 +1,283 @@
...
+	 * Enables smooth scrolling and hash updates for in-page links.
+	 */
+	@Prop() smoothScroll?: boolean = true;
+
+	/**
+	 * Language used for translated defaults.
+	 */
+	@Prop({ mutable: true }) language?: Language;
+
+	@State() translations: any = translations;
```
smorrisods: This is still typed as `any`, and neither `translations` nor `hasDefaultSlotContent` below have JSDocs. The thread on this is marked resolved with "Done in a68b6230," but tracing through that commit, `@State() translations: any = translations;` is unchanged — same on the current branch tip. Since `translations` is just the imported JSON module, `typeof translations` gets us a real type:

```suggestion
	/**
	 * Localized string translations for this component.
	 */
	@State() translations: typeof translations = translations;

	/**
	 * Whether the default slot currently has rendered navigation item content.
	 */
	@State() private hasDefaultSlotContent = false;
```

(Verified via `git show a68b6230:...ontario-in-page-navigation.tsx` — the `any` type was never actually changed despite the reply.)

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
```diff
@@ -0,0 +1,283 @@
...
+		const slotElement = event.target as HTMLSlotElement;
+		this.hasDefaultSlotContent = this.hasSlotContent(slotElement);
+		this.warnIfMissingItems();
+	};
+
+	/**
+	 * Resolves a hash link to the target element in the current document.
+	 */
+	private getAnchorTarget(href: string): HTMLElement | null {
+		if (!href || !href.startsWith('#')) {
```
smorrisods: We check `!href || !href.startsWith('#')` in two places to decide whether a link is an in-page anchor: here in `getAnchorTarget` and again in `handleLinkClick` below (line 195). Tracing it through, `handleLinkClick` already calls `getAnchorTarget(href)` right after its own check, and `getAnchorTarget` returns `null` for anything that isn't `#`-prefixed anyway — so the check at line 195 doesn't change behaviour, it's just a duplicate of this one. Since this exact string check carries meaning ("is this a hash anchor"), let's pull it into a small named helper so both call sites stay in sync if the definition ever changes:

```ts
private isAnchorHref(href?: string | null): href is string {
	return !!href && href.startsWith('#');
}
```

Then `getAnchorTarget` and `handleLinkClick` can both call `this.isAnchorHref(href)` instead of repeating the raw string check.

(Third occurrence at `ontario-in-page-navigation.tsx:258`, `skipLinkHref`, is a normalization — not a validation — check, so it's a looser fit for the helper; mentioned for awareness but not required to change.)

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/ontario-in-page-navigation.tsx
```diff
@@ -0,0 +1,283 @@
+import { Component, Prop, State, Watch, Listen, Element, h } from '@stencil/core';
+
+import { Language } from '../../utils/common/language-types';
+import { HeaderLanguageToggleEventDetails } from '../../utils/events/common-events.interface';
+import { validateLanguage, validateValueAgainstArray } from '../../utils/validation/validation-functions';
+import { ConsoleMessageClass } from '../../utils/console-message/console-message';
+import translations from '../../translations/global.i18n.json';
+
+export type OntarioInPageNavigationHeadingLevel = 'h2' | 'h3' | 'h4';
```
smorrisods: `OntarioInPageNavigationHeadingLevel` is declared as a literal union here (`'h2' | 'h3' | 'h4'`), but then the same three values get re-declared as a separate array literal in `validateHeadingLevel` (line 94, `const allowedValues: OntarioInPageNavigationHeadingLevel[] = ['h2', 'h3', 'h4'];`) — two independent lists that have to be kept in sync by hand.

We already have a pattern for this in the library that avoids the duplication: `ontario-badge.types.ts` and `ontario-page-alert.types.ts` both define a single `as const` array as the source of truth and derive the type from it:

```ts
export const PageAlertTypes = ['informational', 'warning', 'success', 'error'] as const;
export type PageAlertType = (typeof PageAlertTypes)[number];
```

Let's follow the same pattern here so there's only one place these values live:

```suggestion
export const ontarioInPageNavigationHeadingLevels = ['h2', 'h3', 'h4'] as const;
export type OntarioInPageNavigationHeadingLevel = (typeof ontarioInPageNavigationHeadingLevels)[number];
```

and then in `validateHeadingLevel`:

```ts
const allowedValues = ontarioInPageNavigationHeadingLevels;
```

(The `renderHeading` switch at lines 231–235 also hardcodes `'h3'`/`'h4'`/`'h2'`, but that's a normal exhaustive switch over the union rather than a re-declared list, so it's a much looser fit for this — not asking for a change there, just noting it for completeness.)

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/readme.md
```diff
@@ -0,0 +1,130 @@
...
+
+```mdx-code-block
+</TabItem>
+</Tabs>
+```
+
+The canonical authoring pattern in this branch is standard slotted `<li><a></a></li>` items. Once `ontario-in-page-navigation-item` is merged, this documentation will include both patterns and recommend component items for consistency.
```
smorrisods: > The canonical authoring pattern in this branch is standard slotted `<li><a></a></li>` items. Once `ontario-in-page-navigation-item` is merged, this documentation will include both patterns and recommend component items for consistency.

This reads more like a reply to a review comment than documentation — "in this branch" and "once merged" describe git/PR state, not what the component does today, and it'll likely go stale once the item component actually merges. Maybe worth rephrasing to just describe the current pattern:

```suggestion
Slotted list items (`<li><a href="#section-id">Label</a></li>`) are the standard way to author navigation links.
```

smorrisods (changes requested):
Really appreciate the turnaround on the last two rounds — the landmark fix (dropping `role="navigation"` off the `ol`), the named SCSS variables replacing the magic numbers, and the component-level JSDoc all read cleanly now. A few things left to tidy up before this merges, flagged inline.

Small aside, not blocking: the new demo block in `index.html` uses bare `href="#"` for a few placeholder items (`Eligibility`, `Program guide`, `Overview`, etc.) with no matching section `id`, so clicking them jumps to page top per the behaviour we discussed on #257 — though that's consistent with existing placeholder links elsewhere in this file, so not new to this PR.

---
🤖 AI-drafted, Scott-flavoured. No warranty on the jokes, full warranty on the feedback.

smorrisods (approved):
LGTM 🚀

---
🤖 AI-drafted, Scott-flavoured. No warranty on the jokes, full warranty on the feedback.

PR #257, opened 2026-06-05
DS-2665: Implement `ontario-in-page-navigation-item` child component

muntalee:
Implements `ontario-in-page-navigation-item` as a Stencil web component. This is the child component to the `ontario-in-page-navigation` host added in DS-2663.

- `ontario-in-page-navigation-item` component implementation (TSX + SCSS)
- `label` and `href` required props with validation and link rendering
- `isCurrent` prop to expose current section state through `aria-current`
- Slot support for custom item content with prop fallback behavior
- `language` prop with app language event support
- Console warnings when required props are missing/invalid
- Minimal unit test coverage for item behavior
- Item-focused readme examples and usage guidance
- Regenerated React and Angular output targets
- E2E tests
- Accessibility tests

## Out of scope
- React and Angular hand-authored wrappers

## Dependency note for PR 2:
This PR depends on #256 and should be reviewed and merged after it.


packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/test/ontario-in-page-navigation-item.e2e.ts
```diff
@@ -0,0 +1,125 @@
...
+	const expectNoAxeViolations = async (page: E2EPage, selector: string) => {
+		const accessibilityScanResults = await new AxeBuilder({ page }).include(selector).analyze();
+		expect(accessibilityScanResults.violations).toHaveLength(0);
+	};
+
+	test('renders and is hydrated with expected link content', async () => {
+		await expect(host).toBeAttached();
+		await expect(host).toHaveClass(/hydrated/);
+
+		const link = host.locator('a.ontario-page-navigation-item__link');
```
erincarrieresontario: we are re-declaring this quite a few times, we could define it once like this: 
`const getLink = () => host.locator('a.ontario-page-navigation-item__link');`
and then call: 
`await expect(getLink()).toHaveText('Section 1');`

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/test/ontario-in-page-navigation-item.e2e.ts
```diff
@@ -0,0 +1,125 @@
...
+		await expect(link).toHaveAttribute('href', '#section-1');
+		await expect(link).toHaveText('Section 1');
+	});
+
+	test('supports keyboard focus and Enter navigation', async ({ page }) => {
+		await host.evaluate((el: Element) => {
+			(el.shadowRoot?.querySelector('a.ontario-page-navigation-item__link') as HTMLAnchorElement | null)?.focus();
+		});
+
+		const focusedHref = await host.evaluate(
```
erincarrieresontario: This could instead be something like: 
`await expect(getLink()).toBeFocused();`

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/test/ontario-in-page-navigation-item.e2e.ts
```diff
@@ -0,0 +1,125 @@
...
+	test('visual regression: focus-visible state on item link', async ({ page }) => {
+		await host.evaluate((el: Element) => {
+			(el.shadowRoot?.querySelector('a.ontario-page-navigation-item__link') as HTMLAnchorElement | null)?.focus();
+		});
+		await page.waitForChanges();
+
+		const screenshot = await host.screenshot();
+		expect(screenshot.byteLength).toBeGreaterThan(0);
+	});
+});
```
erincarrieresontario, line 100: Could add some tests for missing href & empty label

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/test/ontario-in-page-navigation-item.spec.tsx
```diff
@@ -0,0 +1,102 @@
...
+		expect(link?.getAttribute('href')).toBe('#eligibility');
+	});
+
+	it('warns when label validation runs without slot content and label is missing', async () => {
+		const page = await newSpecPage({
+			components: [OntarioInPageNavigationItem],
+			html: '<ontario-in-page-navigation-item href="#section"></ontario-in-page-navigation-item>',
+		});
+
+		const instance = page.rootInstance as any;
```
erincarrieresontario: Should be : 
`const instance = page.rootInstance as OntarioInPageNavigationItem;`

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.scss
```diff
@@ -0,0 +1,29 @@
+@use 'pkg:@ongov/ontario-design-system-global-styles/styles/scss/1-variables/spacing.variables' as spacing;
+@use 'pkg:@ongov/ontario-design-system-global-styles/styles/scss/1-variables/fonts.variables' as fonts;
+@use 'pkg:@ongov/ontario-design-system-global-styles/styles/scss/1-variables/font-weights.variables' as fontWeights;
+@use 'pkg:@ongov/ontario-design-system-global-styles/styles/scss/2-tools/functions/global.functions' as globalFunctions;
+
+:host {
+	display: list-item;
+	list-style-type: disc;
+	padding-bottom: spacing.$spacing-4 + globalFunctions.px-to-rem(4);
```
erincarrieresontario: This mixes a token with a computed value, which can become inconsistent if tokens change scale.
Either keep everything in tokens or do something like: 
`padding-bottom: calc(#{spacing.$spacing-4} + #{globalFunctions.px-to-rem(4)});`

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.tsx
```diff
@@ -0,0 +1,136 @@
...
+		const defaultSlot = this.host.shadowRoot?.querySelector('slot:not([name])') as HTMLSlotElement | null;
+		this.hasDefaultSlotContent = this.hasSlotContent(defaultSlot || undefined);
+		this.language = validateLanguage(this.language);
+		this.validateLabel();
+		this.validateHref();
+	}
+
+	render() {
+		return (
+			<Host class="ontario-page-navigation-list__item" role="listitem">
```
erincarrieresontario: Since the component is called `ontario-in-page-navigation-item`, this should probably be the base name for the classes on this component. There are some improvements that could be done on the naming conventions in this 👍

erincarrieresontario (changes requested):
Looks good @muntalee , let's just do some adjusting to the class names for this component 👍

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/test/ontario-in-page-navigation-item.e2e.ts
muntalee: Thats great! I added it in ff6053ea.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/test/ontario-in-page-navigation-item.e2e.ts
muntalee: Valid, I changed it in e417429c.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/test/ontario-in-page-navigation-item.e2e.ts
muntalee: Oversight on my part, I added it in 7e56858c

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/test/ontario-in-page-navigation-item.spec.tsx
muntalee: Done in 9134e537!

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.scss
muntalee: Thats a good idea, I'll add the calc in 7a2f48d3 for now and will come back in the design tokens.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.tsx
muntalee: Valid point for consistency, refactored it in 3059e231.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.tsx
```diff
@@ -0,0 +1,136 @@
...
+		this.validateHref();
+	}
+
+	render() {
+		return (
+			<Host class="ontario-in-page-navigation-item" role="listitem">
+				<slot onSlotchange={this.handleDefaultSlotChange}>
+					<a
+						class="ontario-in-page-navigation-item__link"
+						href={this.href?.trim() || '#'}
```
smorrisods: When href is missing we fall back to href="#" here. Tracing it through, the host's getAnchorTarget slices off the # and gets an empty targetId, so it returns null and preventDefault never gets called (see ontario-in-page-navigation.tsx around line 190). That means clicking a misconfigured item scrolls the page to the top and appends # to the URL, which is a confusing side effect on top of the console warning. Could we omit the href attribute entirely in this case, or have the host swallow clicks on a bare # href?

smorrisods (changes requested):
Good stuff overall, the role="list"/"listitem" fix for the custom elements is a nice catch. Two things I'd like addressed before this merges.

The item click bug is called out inline below. The other one doesn't have a diff to attach a comment to: components.d.ts wasn't regenerated for either ontario-in-page-navigation or ontario-in-page-navigation-item, but the React and Angular wrappers already reference their types (like HTMLOntarioInPageNavigationItemElement), and the new e2e test casts to that type directly. That'll fail to resolve until we run the stencil build/docs step again.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.tsx
muntalee: Thank you for the feedback, I fixed both points. 

I removed the `href="#"` fallback from `ontario-in-page-navigation-item` in babc24c when href is missing, so a misconfigured item now warns without jumping the page to the top or appending # to the URL. The Stencil build/docs generation has also been done in e2467ddf, so `components.d.ts` now includes the host and item component types used by the wrappers and typed tests.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.tsx
```diff
@@ -0,0 +1,138 @@
...
+	 * Marks the current/active section.
+	 */
+	@Prop() isCurrent?: boolean = false;
+
+	/**
+	 * Language used if localized text is required in the future.
+	 */
+	@Prop({ mutable: true }) language?: Language;
+
+	@State() private hasDefaultSlotContent = false;
```
smorrisods, line 44: This should have a JSDoc comment.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+	private hasSlotContent(slotElement?: HTMLSlotElement): boolean {
```
smorrisods, line 49: In fact all these items should have a JSDoc comment

smorrisods (changes requested):
Good work here! Some minor comments in this follow up review.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.tsx
muntalee: Apologies for the oversight, added them in e818d23

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.tsx
```diff
@@ -0,0 +1,156 @@
...
+import { ConsoleMessageClass } from '../../utils/console-message/console-message';
+import { Language } from '../../utils/common/language-types';
+import { validateLanguage } from '../../utils/validation/validation-functions';
+
+@Component({
+	tag: 'ontario-in-page-navigation-item',
+	styleUrl: 'ontario-in-page-navigation-item.scss',
+	shadow: true,
+})
+export class OntarioInPageNavigationItem {
```
smorrisods, line 18: The component is missing a JSDoc description, which we're trying to introduce across all components to better improve the documentation. Checkout the descriptions from other components for more detail.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/readme.md
```diff
@@ -0,0 +1,183 @@
...
+<ontario-in-page-navigation-item
+	label="Eligibility"
+	href="#eligibility"
+	isCurrent="true"
+></ontario-in-page-navigation-item>
+```
+
+When implemented, this should apply `aria-current="true"`.
+
+1. Slot override pattern
```
smorrisods, line 138: The lines below each markdown list item should be indented so the ordered list can render.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/readme.md
```diff
@@ -0,0 +1,183 @@
...
+<!-- Auto Generated Below -->
+
+## Properties
+
+| Property    | Attribute    | Description                                                | Type                        | Default     |
+| ----------- | ------------ | ---------------------------------------------------------- | --------------------------- | ----------- |
+| `href`      | `href`       | In-page anchor target, for example #eligibility.           | `string \| undefined`       | `undefined` |
+| `isCurrent` | `is-current` | Marks the current/active section.                          | `boolean \| undefined`      | `false`     |
+| `label`     | `label`      | Link label for the in-page navigation item.                | `string \| undefined`       | `undefined` |
+| `language`  | `language`   | Language used if localized text is required in the future. | `"en" \| "fr" \| undefined` | `undefined` |
```
smorrisods, line 185: There don't appear to be any properties that are required, what does the component do when the user doesn't give it anything? Does it gracefully render nothing?

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/test/ontario-in-page-navigation.e2e.ts
```diff
@@ -1,13 +1,260 @@
...
+		expect(accessibilityScanResults.violations).toHaveLength(0);
+	};
+
+	const clickItemLink = async (page: E2EPage, index: number) => {
+		await page
+			.locator('ontario-in-page-navigation-item')
+			.nth(index)
+			.evaluate((el: Element) => {
+				const anchor = el.shadowRoot?.querySelector(
+					'a.ontario-in-page-navigation-item__link',
```
smorrisods: We're re-declaring this selector 4 times across separate tests in this file (`clickItemLink`, the focus tests, etc.) - the same thing we just cleaned up in the item e2e tests. We could pull it into a shared helper once:

```ts
const getItemLink = (el: Element) =>
	el.shadowRoot?.querySelector('a.ontario-in-page-navigation-item__link') as HTMLAnchorElement | null;
```

and reuse it here and in the other tests below to keep both test files consistent.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/test/ontario-in-page-navigation.e2e.ts
```diff
@@ -1,13 +1,260 @@
...
+
+		await firstItem.evaluate((el: Element) => {
+			(el.shadowRoot?.querySelector('a.ontario-in-page-navigation-item__link') as HTMLAnchorElement | null)?.focus();
+		});
+		await page.waitForChanges();
+
+		const screenshot = await host.screenshot();
+		expect(screenshot.byteLength).toBeGreaterThan(0);
 	});
 });
```
smorrisods: There are Visual Regression tests still in the E2E tests, now we have the infrastructure for VRTs these can be ported to a `.vrt.ts` file.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.scss
```diff
@@ -0,0 +1,29 @@
+@use 'pkg:@ongov/ontario-design-system-global-styles/styles/scss/1-variables/spacing.variables' as spacing;
+@use 'pkg:@ongov/ontario-design-system-global-styles/styles/scss/1-variables/fonts.variables' as fonts;
+@use 'pkg:@ongov/ontario-design-system-global-styles/styles/scss/1-variables/font-weights.variables' as fontWeights;
+@use 'pkg:@ongov/ontario-design-system-global-styles/styles/scss/2-tools/functions/global.functions' as globalFunctions;
+
+:host {
+	display: list-item;
+	list-style-type: disc;
+	padding-bottom: calc(#{spacing.$spacing-4} + #{globalFunctions.px-to-rem(4)});
```
smorrisods: Let's drop the `calc()` and use plain SCSS addition, since both operands here are compile-time SCSS values (not CSS custom properties). `calc()` forces the browser to evaluate this expression at layout/paint time on every render, when Sass can already resolve it to a single static rem value at build time - same pattern already used in `ontario-in-page-navigation.scss`:

```scss
padding-bottom: spacing.$spacing-4 + globalFunctions.px-to-rem(4);
```

Do we have a spacing var that will be `4px` already defined that can be used instead of `px-to-rem`?

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.scss
```diff
@@ -0,0 +1,29 @@
...
+}
+
+:host(:last-child) {
+	padding-bottom: spacing.$spacing-0;
+}
+
+.ontario-in-page-navigation-item__link {
+	font-family: fonts.$ontario-font-open-sans;
+	font-size: globalFunctions.px-to-rem(18);
+	line-height: globalFunctions.px-to-rem(25.6);
```
smorrisods: Let's turn these into named variables since they carry meaning and are magic numbers.

smorrisods (changes requested):
@muntalee, great work! I've finished my full review, I've flagged a few items.  If you have any questions let me know or reply to the comments.

🧶 _🐈_

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.tsx
muntalee: Added component class JSDoc in cd368cec.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/readme.md
muntalee: Fixed markdown list indentation and rendering in aef8eb18.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/readme.md
muntalee: Documented fallback behaviour when props are missing in aef8eb18.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/test/ontario-in-page-navigation.e2e.ts
muntalee: Refactored to a shared link locator helper in 4de4c126.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/test/ontario-in-page-navigation.e2e.ts
muntalee: True! Migrated visual assertions into a dedicated VRT suite in 4de4c126.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.scss
muntalee: Replaced calc and used tokenized compile-time SCSS values in 5b4a43f6.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.scss
muntalee: Moved to named style variables in 5b4a43f6.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.tsx
```diff
@@ -0,0 +1,159 @@
...
+	render() {
+		const resolvedHref = this.href?.trim();
+
+		return (
+			<Host class="ontario-in-page-navigation-item" role="listitem">
+				<slot onSlotchange={this.handleDefaultSlotChange}>
+					<a
+						class="ontario-in-page-navigation-item__link"
+						href={resolvedHref || undefined}
+						aria-current={this.isCurrent ? 'true' : undefined}
```
erincarrieresontario: Should this use aria-current="location" instead of "true"? Since this component is used for in-page navigation, location may better describe that the active item corresponds to the user's current section within the page.

erincarrieresontario (approved):
Looks good @muntalee , just one suggestion!

erincarrieresontario (approved):
Looks good @muntalee , just one suggestion!

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.tsx
muntalee: I fixed it up in 1832428 as it was more semantically appropriate choice.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.tsx
```diff
@@ -0,0 +1,159 @@
+import { Component, Prop, h, Host, Watch, State, Element } from '@stencil/core';
+
+import { ConsoleMessageClass } from '../../utils/console-message/console-message';
+import { Language } from '../../utils/common/language-types';
+import { validateLanguage } from '../../utils/validation/validation-functions';
+
+@Component({
```
smorrisods, line 13: This component is missing JSDocs at the Component class level that matches our other components.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.tsx
```diff
@@ -0,0 +1,159 @@
...
+	tag: 'ontario-in-page-navigation-item',
+	styleUrl: 'ontario-in-page-navigation-item.scss',
+	shadow: true,
+})
+/**
+ * A single list item link used inside `ontario-in-page-navigation`.
+ */
```
smorrisods: Our other components have the JSDocs above the `@Component` directive.  See [`OntarioButton`](https://github.com/ongov/ontario-design-system/blob/d21a430f92640f9a0297b84c7268ee07936e3862/packages/ontario-design-system-component-library/src/components/ontario-button/ontario-button.tsx#L10) for more details.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/readme.md
```diff
@@ -0,0 +1,181 @@
...
+		{label: 'Angular', value: 'angular'},
+	]}
+	groupId="framework"
+	queryString="framework">
+<TabItem value="html">
+```
+
+```html
+<ontario-in-page-navigation-item label="Eligibility" href="#eligibility"></ontario-in-page-navigation-item>
+```
```
smorrisods, line 44: Since you've used an indent on line `27` these example code blocks might need to be indented as well.  Maybe try running it through docusaurus to see what they look like

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/test/ontario-in-page-navigation-item.e2e.ts
```diff
@@ -0,0 +1,114 @@
...
+		await expect(host.locator('a.custom-item-link')).toHaveText('Custom section');
+	});
+
+	test('has no axe violations', async ({ page }) => {
+		await expectNoAxeViolations(page, 'ontario-in-page-navigation-item');
+	});
+
+	test('visual regression: default item', async () => {
+		const screenshot = await host.screenshot();
+		expect(screenshot.byteLength).toBeGreaterThan(0);
```
smorrisods: These are checking that the screenshot exists, not if it is correct for not.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/test/ontario-in-page-navigation.vrt.ts
```diff
@@ -0,0 +1,60 @@
...
+		host = page.locator('ontario-in-page-navigation').first();
+	};
+
+	test.beforeEach(async ({ page }) => {
+		await setHostContent(page);
+	});
+
+	test('default variant', async () => {
+		await expect(host).toHaveClass(/hydrated/);
+		await expect(host).toHaveScreenshot('ontario-in-page-navigation-default.png', screenshotOptions);
```
smorrisods: Let's use the autogenerated names rather than forcing our own so we don't have to maintain them.

This applies to each instance of `toHaveScreenshot`

smorrisods (changes requested):
It's coming along nicely, a few observations to clean things up.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.tsx
muntalee: Moved the JSDoc block above the `@Component` decorator in 87ff4d21

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/ontario-in-page-navigation-item.tsx
muntalee: Fixed in 87ff4d21

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/test/ontario-in-page-navigation-item.e2e.ts
muntalee: Got that fixed up in df476fad with all new VRT suite for the in page nav item.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation/test/ontario-in-page-navigation.vrt.ts
muntalee: Gotchu, removed the explicit `.png` filename arguments from all three `toHaveScreenshot()` calls and regenerated the baselines with their autogenerated name in e2e4554e.

packages/ontario-design-system-component-library/src/components/ontario-in-page-navigation-item/readme.md
smorrisods: The examples in this file are wrapped in a numbered list (`1. Standard item`, `1. Item with long label`, etc.) with the description line indented underneath, but the `mdx-code-block`/`html`/`tsx` fences that follow sit at column 0 — so they don't actually nest inside the list item.

Every other component readme that uses `Tabs`/`TabItem` (`ontario-button`, `ontario-badge`, `ontario-accordion`, etc.) skips the list wrapper entirely — examples are just flat prose followed directly by column-0 fences. Let's follow that same convention here instead of trying to fix the indentation:

```suggestion
Example of a standard item, demonstrating the basic required configuration:
```

Same change applies to the other three examples further down (`Item with long label`, `Current (active) item`, `Slot override pattern`) — drop the `1. ` list marker and indentation, keep the description as flat prose above the fences.

smorrisods (approved):
LGTM 🚀

PR #258, opened 2026-06-05
DS-1919: Add missing checkbox and radio margin override in last form group

muntalee:
Adds the missing checkbox and radio `margin-bottom` override for the last `.ontario-form-group`, matching the Fractal behavior described in DS-1919.

- Add the missing `.ontario-checkboxes` and `.ontario-radios` `margin-bottom: 0` override inside `.ontario-form-group:last-of-type`
- Add `:host-context(.ontario-form-group:last-of-type)` handling in `ontario-checkboxes` so the override applies through Shadow DOM
- Add matching `:host-context(.ontario-form-group:last-of-type)` handling in `ontario-radio-buttons`
- Keep the fix scoped to the existing form-group behavior without changing other spacing rules
- Align web component behavior with the Fractal reference and Angular POC expectation

### DIFF
--- a/packages/ontario-design-system-component-library/src/components/ontario-checkbox/ontario-checkboxes.scss
+++ b/packages/ontario-design-system-component-library/src/components/ontario-checkbox/ontario-checkboxes.scss
@@ -26,6 +26,10 @@ $ontario-checkbox-box-shadow-outline: globalFunctions.px-to-rem(4);
 	max-width: globalVariables.$standard-width;
 }
 
+:host-context(.ontario-form-group:last-of-type) .ontario-checkboxes {
+	margin-bottom: spacing.$spacing-0;
+}
+
 .ontario-checkboxes__item {
 	position: relative;
 	margin: 0 0 spacing.$spacing-4;
--- a/packages/ontario-design-system-global-styles/src/styles/scss/5-layout/_form-group.layout.scss
+++ b/packages/ontario-design-system-global-styles/src/styles/scss/5-layout/_form-group.layout.scss
@@ -4,4 +4,9 @@
 @warn "The `ontario-form-group` class will be deprecated and replaced with a vertical spacing component in future.";
 .ontario-form-group:last-of-type {
 	margin-bottom: spacing.$spacing-8;
+
+	.ontario-checkboxes,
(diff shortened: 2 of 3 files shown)


erincarrieresontario (approved):
Looks good to me 👍

packages/ontario-design-system-global-styles/src/styles/scss/5-layout/_form-group.layout.scss
```diff
@@ -4,4 +4,9 @@
 @warn "The `ontario-form-group` class will be deprecated and replaced with a vertical spacing component in future.";
 .ontario-form-group:last-of-type {
 	margin-bottom: spacing.$spacing-8;
+
+	.ontario-checkboxes,
+	.ontario-radios {
+		margin-bottom: spacing.$spacing-0;
+	}
```
smorrisods, line 11: We have to be careful here, we're redone how margins work and I don't think this is the solution the holistically solves the bug.  The bug is now over 2 years old and we added vertical spacing to better handle form element spacing.

packages/ontario-design-system-component-library/src/components/ontario-checkbox/ontario-checkboxes.scss
```diff
@@ -26,6 +26,10 @@ $ontario-checkbox-box-shadow-outline: globalFunctions.px-to-rem(4);
 	max-width: globalVariables.$standard-width;
 }
 
+:host-context(.ontario-form-group:last-of-type) .ontario-checkboxes {
+	margin-bottom: spacing.$spacing-0;
+}
+
```
smorrisods, line 32: The form elements shouldn't be handling any spacing themselves. This is part of the vertical spacing work the @MeaghanCarrieres did.

smorrisods (changes requested):
Good work @muntalee.  However, I think this bug is no longer relevant since we changed how spacing worked with form elements since it was created.  The form elements shouldn't be aware of their margins between each other as this is handled by the vertical spacing wrapper we've added.  @erincarrieresontario, I'm curious your thoughts.

smorrisods:
DS-1919 is no longer relevant due to our work with vertical form spacing.

PR #264, opened 2026-06-15
Migrate HTML generator tests from Jest to Vitest

smorrisods:
## Summary

Migrates the standalone `ontario-design-system-component-html-generator` test suite from **Jest/ts-jest** to **Vitest**.

- Replaces `jest.config.mjs` with `vitest.config.ts` (node environment).
- Swaps the Jest devDependencies (`@jest/globals`, `@types/jest`, `jest`, `jest-environment-node`, `ts-jest`) for `vitest`.
- Uses explicit `vitest` imports in `utils.test.ts` (no global typings needed).
- Excludes `*.test.ts` from the `tsconfig` build so test files no longer ship in `dist/`.
- `test` script now runs `vitest run`; adds a `test:watch` script.

All **33** existing tests pass unchanged under Vitest, and `pnpm run build` produces a clean `dist/` with no test files.

## Motivation

Stencil's built-in test runner is **deprecated** (removed in Stencil v5), and Stencil's `getJestFacade()` only supports Jest 27/28/29 — so Jest is a dead end in this repo and blocks timely security updates (e.g. the Jest v30 bump cannot land). Moving to Vitest gives this package a forward-facing, maintained test runner that we can keep patched.

This is the first of a small series of PRs replacing Jest with Vitest across the monorepo. The `html-generator` is a self-contained, low-risk starting point.

## Context

- This package used a standalone Jest setup (not Stencil's test runner), so the migration is isolated and does not affect any other package.
- The `pnpm-lock.yaml` change is the real dependency delta only (the lockfile is re-formatted with Prettier to match the committed style).

## Testing

- `pnpm run test` → 33 passed.
- `pnpm run build` → succeeds; verified no `*.test.*` files in `dist/`.

PR #265, opened 2026-06-16
Migrate component library unit tests from Jest to Vitest

smorrisods:
## Summary

Migrates the Ontario Design System component library's Stencil **unit tests** from Jest to **Vitest**, continuing the broader testing migration (the HTML generator migration is in #264). Playwright end-to-end tests are unchanged.

## Motivation

Stencil's built-in Jest runner is being retired in favour of Vitest via the official `@stencil/vitest` integration. Standardizing on Vitest gives us a faster, actively maintained runner and a single test toolchain across packages.

## What changed

- **Runner:** Adopt `@stencil/vitest` with a dist-loader setup.
  - `stencil.config.test.ts` builds only the lazy-loader `dist` output target (avoids the heavier, side-effecting React/Angular wrapper generation and docs targets during test builds).
  - `vitest-setup.ts` imports that built bundle so the custom elements are registered for the spec run.
  - `vitest.config.ts` uses the Stencil `stencil` environment (mock-doc), so `toEqualHtml` output matches the previous `newSpecPage` format.
- **Typing:** `tsconfig.spec.json` provides Vitest globals + bundler resolution for editor support; spec and test-helper files are excluded from the Stencil production build (they are validated by running Vitest, mirroring the old transpile-only Jest behaviour).
- **Specs:** Migrated all spec files — `newSpecPage` → `render`, `page.rootInstance` → `page.instance`, `jest.*` → `vi.*`, `SpecPage` → `RenderResult` — and regenerated snapshots and inline `toEqualHtml` expectations to match the fully hydrated render output.
- **Scripts:** `test:unit` now builds the test config then runs `vitest run`; `test:e2e` (Playwright) is unchanged.
- **Cleanup:** Removed the Jest devDependencies, the Jest `testing` block in `stencil.config.ts`, the Jest SVG transformer, and switched `tsconfig` types off `jest`.

## Testing

- `pnpm run test:unit` → **35 test files, 202 passed, 2 skipped** (the 2 skips pre-existed on `develop`).
- Playwright e2e suite unaffected.


muntalee (approved):
LGTM 🐐

PR #267, opened 2026-06-16
chore: export the taskStatuses enum

erincarrieresontario:
Fix: Refactored the task status API to use a string-union type pattern instead of special-case enum exports. Consumers now pass task status as string literals (e.g., [taskStatus="completed"]) with full type safety, eliminating confusion about the intended API.


packages/ontario-design-system-component-library/src/index.ts
```diff
@@ -1,2 +1,3 @@
 // export { Components, JSX } from './components';
 export * from './components';
+export { TaskStatuses } from './utils/common/task-statuses.enum';
```
smorrisods: I'm curious why this particular enum is being exported here vs. just in the regular way as part of the component?

packages/ontario-design-system-component-library-react/src/task-statuses.ts
```diff
@@ -0,0 +1,8 @@
+export const TaskStatuses = {
+	NotStarted: 'notStarted',
+	InProgress: 'inProgress',
+	Completed: 'completed',
+	CannotStartYet: 'cannotStartYet',
+	Error: 'error',
+	Optional: 'optional',
+} as const;
```
smorrisods: Why are we updating the React library for this particular enum?

smorrisods (changes requested):
This change does not align with now we're exporting other emums and values in any of the components.  It adds more maintenance overhead to the React library for this special case.

Let's chat about this and see what is up and how to address it.

packages/ontario-design-system-component-library/src/components.d.ts
```diff
@@ -4,9019 +4,9341 @@
...
+	HeaderMenuToggleDetail,
+	LanguageToggleOptions,
+	OntarioHeaderType,
+} from './components/ontario-header/ontario-header.interface';
+export { IconColour, IconSize } from './components/ontario-icon/icon.types';
+export { HeaderLanguageToggleEventDetails } from './utils/events/common-events.interface';
+export { PageAlertType } from './components/ontario-page-alert/ontario-page-alert.interface';
+export { RadioOption } from './components/ontario-radio-buttons/radio-option.interface';
+export { TableColumnOptions, TableRowOptions } from './components/ontario-table/table.interface';
+export { TaskStatus } from './utils/common/task-statuses.enum';
```
smorrisods: This line is exporting `TaskStatus`, we shouldn't have to manually export it.

packages/ontario-design-system-component-library/src/index.ts
```diff
@@ -1,2 +1,3 @@
 // export { Components, JSX } from './components';
 export * from './components';
+export { TaskStatus } from './utils/common/task-statuses.enum';
```
smorrisods: This is already exported via `packages/ontario-design-system-component-library/src/components.d.ts`

smorrisods (changes requested):

smorrisods (approved):
LGTM 🚀

PR #266, opened 2026-06-16
Remove vestigial Jest config from React component library

smorrisods:
## Summary

Removes the vestigial Jest configuration and Jest dependencies from the React component library wrapper. This package is auto-generated by the Stencil React output target and contains no tests, no `jest.setup.js`, and no `test` script — the Jest config block referenced a setup file that does not exist.

This is part of the broader Jest → Vitest migration (see #264, #265). Since the React wrapper is purely generated glue code with no test suite, it uses no test runner ("vitest or nothing").

## Changes

- Remove the dead `jest` configuration block from `package.json` (it referenced a non-existent `jest.setup.js`).
- Remove Jest devDependencies: `jest`, `@types/jest`, `@testing-library/jest-dom`.
- Remove the unused `jest` entry from the `types` array in `tsconfig.json`.
- Update `pnpm-lock.yaml` accordingly.

## Motivation

The upstream `@stencil/react-output-target` example keeps wrapper integration tests in a separate area, not inside the generated wrapper package. There are no tests to migrate here, so the cleanest path is to drop the unused Jest tooling rather than introduce a test runner for generated code.

## Testing

- No test files exist in this package; nothing to run.
- `pnpm install` succeeds and the Jest tree is removed from the lockfile.

PR #268, opened 2026-06-16
Remove vestigial Jest dependencies from React demo app

smorrisods:
## Summary

Removes vestigial Jest dependencies from the React demo (PoC) app. The app has no test runner, no test files, and no test imports — its `test` script is a placeholder (`echo "No tests configured"`).

This is part of the broader Jest → Vitest migration (see #264, #265, #266).

## Changes

- Remove unused Jest dependencies from `packages/app-react/package.json`:
  - `@testing-library/jest-dom`
  - `@types/jest`
- Update `pnpm-lock.yaml` accordingly.

## Notes

The app also lists `@testing-library/react` and `@testing-library/user-event`, which are currently unused as well (no test runner). They are framework-agnostic (not Jest-specific), so they are left in place for now and can be removed separately if the app continues to ship without tests.

## Testing

- No test files exist in this package; nothing to run.
- `pnpm install` succeeds with the Jest packages removed from the lockfile.


muntalee (approved):
LGTM 🐐 We could consider removing `@testing-library/react` and `@testing-library/user-event`, as their currently unused, but no problems besides that.

PR #259, opened 2026-06-14
chore(deps): update dependency @angular/core to v22

smorrisods:
This PR contains the following updates:

| Package | Change | Age | Confidence |
|---|---|---|---|
| [@angular/core](https://redirect.github.com/angular/angular) ([source](https://redirect.github.com/angular/angular/tree/HEAD/packages/core)) | [`^15.2.9` -> `^22.0.0`](https://renovatebot.com/diffs/npm/@angular%2fcore/15.2.1/22.0.1) | [age](https://docs.renovatebot.com/merge-confidence/) | [confidence](https://docs.renovatebot.com/merge-confidence/) |

---

### Release Notes

<details>
<summary>angular/angular (@&#8203;angular/core)</summary>

### [`v22.0.1`](https://redirect.github.com/angular/angular/blob/HEAD/CHANGELOG.md#2201-2026-06-10)

[Compare Source](https://redirect.github.com/angular/angular/compare/v22.0.0...v22.0.1)

##### platform-server

- XHR support in `@angular/platform-server` is deprecated. Use standard `fetch` APIs instead.
  (cherry picked from commit [`8446e46`](https://redirect.github.com/angular/angular/commit/8446e46f8bc33bd4419fa7f6106b8d117ca2e099))

##### common

| Commit                                                                                           | Type | Description                                                            |
| ------------------------------------------------------------------------------------------------ | ---- | ---------------------------------------------------------------------- |
| [c4b5fa3c92](https://redirect.github.com/angular/angular/commit/c4b5fa3c9263ac127f5053c5a03dd4b6313659b8) | fix  | escape CSS string-terminating characters in escapeCssUrl               |
| [dfff57ede9](https://redirect.github.com/angular/angular/commit/dfff57ede93dbc51a7eeac3311ff2b1279595ee5) | fix  | Limits date format string length                                       |
| [3c2892c8df](https://redirect.github.com/angular/angular/commit/3c2892c8dffbbbe32940306b53779cc0c4e3f73c) | fix  | prevent prototype pollution in formatDateTime                          |
| [1d87c49f6e](https://redirect.github.com/angular/angular/commit/1d87c49f6ee4aac27146f39ef370a87ba707a2c1) | fix  | use cryptographically secure SHA-256 for transfer cache key generation |

##### compiler

| Commit                                                                                           | Type | Description                                                                    |
| ------------------------------------------------------------------------------------------------ | ---- | ------------------------------------------------------------------------------ |
[...]

PR #269, opened 2026-06-18
Clean up orphan nested manifest in Angular wrapper package

smorrisods:
## Summary

Cleans up the orphaned nested manifest in the Angular wrapper package that has been causing stale, partial Renovate PRs (most recently #259).

## Background

`packages/ontario-design-system-component-library-angular/projects/component-library` is an `ng-packagr` library project. Its `package.json` is only a **build input** — `ng-packagr` reads it and copies it into the build output at `dist/component-library/package.json`. The package we actually publish to npm is the **top-level** `@ongov/ontario-design-system-component-library-angular` manifest, which already declares the correct Angular peer range (`>=18 <23`).

The nested folder is **not** a pnpm workspace member (`pnpm-workspace.yaml` only globs `packages/*`), so lerna/semantic-release never version or update it. It has been frozen at its original proof-of-concept scaffolding (`version: 0.0.1`) for over two years.

## The problem

A leftover npm `package-lock.json` sat next to the nested manifest. Renovate scans the filesystem independently of pnpm's workspace config, so it treated that `package.json` + `package-lock.json` pair as a separate npm project. This caused:

- Stale, inconsistent dependency bumps against dead metadata (e.g. #259 bumped `@angular/core` to `^22` while leaving `@angular/common` at `^15`).
- The `renovate/artifacts` check failing, because Renovate tried to `npm install` to regenerate that orphan lockfile inside a pnpm monorepo.

## Changes

- **Delete** the orphan `projects/component-library/package-lock.json` (npm artifact in a pnpm repo — nothing reads it).
- **Align** the nested manifest's Angular `peerDependencies` with the published top-level range (`>=18 <23`) so the metadata baked into the build output is consistent and no longer contradictory.
- **Add** a Renovate `ignorePaths` entry for the nested manifest so it is no longer dependency-managed (the default `**/node_modules/**` is preserved since specifying `ignorePaths` overrides Renovate's default).

## Follow-up

Supersedes and closes #259.


erincarrieresontario (approved):
LGTM 👍

PR #271, opened 2026-06-18
DS-2671: Implemented new ontario-back-button web component

erincarrieresontario:
This MR introduces a new `ontario-back-button` web component.

The component is navigation-strategy agnostic and supports three modes:
- `history` (default): emits `backClick` and calls `window.history.back()`
- `href`: emits `backClick` and renders as a link when `href` is provided
- `event`: emits `backClick` only (app/router decides navigation)

It also includes bilingual defaults (`Back`/`Retour`), disabled behavior, accessibility handling, and comprehensive unit + E2E test coverage.

### New component implementation
- Added `ontario-back-button` component:
  - `packages/ontario-design-system-component-library/src/components/ontario-back-button/ontario-back-button.tsx`
  - `packages/ontario-design-system-component-library/src/components/ontario-back-button/ontario-back-button.scss`
- Added `BackMode` type: `'history' | 'href' | 'event'`
- Added props:
  - `label?: string`
  - `language?: 'en' | 'fr'`
  - `href?: string`
  - `backMode: BackMode` (default: `history`)
  - `disabled?: boolean` (default: `false`)
- Added event:
  - `backClick` (`MouseEvent | KeyboardEvent`)
- Added runtime validation:
  - warns when `backMode="href"` is used without `href`
  - falls back safely to button rendering in that case
- Added language handling from existing app/header language events.

### Docs and guidance
- Added/expanded component docs:
  - `packages/ontario-design-system-component-library/src/components/ontario-back-button/readme.md`
- Updated component inventory/readme:
  - `packages/ontario-design-system-component-library/src/components/readme.md`
- Updated sample/demo page with back-button section:
  - `packages/ontario-design-system-component-library/src/index.html`
- Added i18n key for label text:
  - `packages/ontario-design-system-component-library/src/translations/global.i18n.json`
  - `backButton.back.en = "Back"`, `backButton.back.fr = "Retour"`

### Generated typings and wrappers
- Stencil generated typings/events:
  - `packages/ontario-design-system-component-library/src/components.d.ts`
- React wrappers (client + server):
  - `packages/ontario-design-system-component-library-react/src/components/components.ts`
  - `packages/ontario-design-system-component-library-react/src/components/components.server.ts`
- Angular proxy + directive registration:
  - `packages/ontario-design-system-component-library-angular/projects/component-library/src/lib/stencil-generated/components.ts`
  - `packages/ontario-design-system-component-library-angular/projects/component-library/src/lib/stencil-generated/index.ts`

### Tests
- Added/expanded unit tests:
  - `packages/ontario-design-system-component-library/src/components/ontario-back-button/test/ontario-back-button.spec.tsx`
- Added E2E tests:
  - `packages/ontario-design-system-component-library/src/components/ontario-back-button/test/ontario-back-button.e2e.ts`
- Coverage includes:
  - rendering and mode behavior (`history`/`href`/`event`)
  - event emission
  - disabled behavior
  - keyboard interaction
  - bilingual label behavior
  - accessibility assertions
  - configuration warning behavior

## Impact

- ✅ New component API surface: `ontario-back-button`
- ✅ React and Angular consumers can use the generated wrappers
- ✅ No intentional breaking changes to existing component APIs
- ℹ️ Includes generated file updates and a large `pnpm-lock.yaml` diff

## Validation / Testing

Implemented test suites for:
- unit/spec behavior
- E2E behavior and interaction scenarios

(If CI/job output is available, link or paste results here.)


packages/ontario-design-system-component-library/src/components/ontario-back-button/ontario-back-button.tsx
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	/**
+	 * Resolves which navigation mode to use at runtime.
+	 *
+	 * If `href` is provided without an explicit `back-mode` attribute,
+	 * this defaults to `href` mode for expected link behaviour.
+	 */
+	private get resolvedBackMode(): BackMode {
+		const hasBackModeAttribute = this.host.hasAttribute('back-mode');
```
smorrisods: I think this might break for React/Angular consumers — both generated wrapper layers set `backMode` via direct property assignment (`node[name] = value`), never `setAttribute`. So `hasAttribute('back-mode')` would report `false` even when a consumer explicitly sets `backMode="event"` (or `"history"`) alongside `href`, silently overriding their choice to `'href'` mode. Can we find another way to detect "was this explicitly set" that doesn't rely on the attribute being present?

packages/ontario-design-system-component-library/src/components/ontario-back-button/ontario-back-button.tsx
```diff
@@ -0,0 +1,222 @@
...
+	 * - `history`: emits event, then calls browser history back.
+	 * - `href`: emits event, then navigates using `href`.
+	 * - `event`: emits event only.
+	 */
+	@Prop() backMode: BackMode = 'history';
+
+	/**
+	 * Disables user interaction.
+	 */
+	@Prop() disabled?: boolean = false;
```
smorrisods, line 69: Curious about this one — `ontario-button` intentionally doesn't expose a `disabled` prop because of our documented policy (keep actions available, guide with validation/error messaging instead of disabling). Is back-button meant to be an exception to that? If so let's document the rationale here like we do over there, otherwise might be worth aligning with the same pattern.

packages/ontario-design-system-component-library/src/components/ontario-back-button/test/ontario-back-button.spec.tsx
```diff
@@ -0,0 +1,241 @@
...
+		it('activates on Enter key press', async () => {
+			const page = await render(`<ontario-back-button back-mode="event"></ontario-back-button>`);
+
+			const eventSpy = vi.fn();
+			page.root?.addEventListener('backClick', eventSpy);
+
+			const button = page.root?.shadowRoot?.querySelector('button') as HTMLButtonElement;
+			const enterEvent = new KeyboardEvent('keydown', { key: 'Enter', bubbles: true });
+			button.dispatchEvent(enterEvent);
+			button.click();
```
smorrisods: This test dispatches a `keydown` event but then unconditionally calls `button.click()` right after — since there's no `onKeyDown` handler in the component, this test would pass identically even if keyboard activation were completely broken (same thing happens in the Space key test below). Might be worth removing the keydown dispatch since the e2e/Playwright test already covers real keyboard behaviour, or reworking this so it actually exercises something the component controls.

smorrisods (changes requested):
Thanks for putting this together, the component itself is in good shape (nice work on the SSR guard and i18n handling) 🙌 Left a few comments below — nothing huge, just want to chat through them before merging.

packages/ontario-design-system-component-library/src/components/ontario-back-button/test/ontario-back-button.e2e.ts
```diff
@@ -0,0 +1,123 @@
...
+	test('renders and hydrates', async () => {
+		await expect(host).toBeAttached();
+		await expect(host).toHaveClass(/hydrated/);
+	});
+
+	test('renders button in history mode by default', async () => {
+		const button = host.locator('button');
+		await expect(button).toBeVisible();
+		await expect(button).toContainText('Back');
+		await expect(host.locator('a')).toHaveCount(0);
```
smorrisods, line 24: What's the check for `<a>` not existing for, to make sure it's a button and not a link?

packages/ontario-design-system-component-library/src/components/ontario-back-button/ontario-back-button.tsx
erincarrieresontario: Yep, good catch on this. I updated it so we’re not relying on on the `hasAttribute('back-mode')` anymore, since React/Angular wrappers set props directly. Now if `backMode` is explicitly set (including property assignment), that value wins even when `href` is present. If `backMode` isn’t set, it falls back to `href` mode when `href`  exists, otherwise `history`. I also added unit and e2e tests for the property-assignment case to make sure wrapper behaviour is covered.

packages/ontario-design-system-component-library/src/components/ontario-back-button/ontario-back-button.tsx
erincarrieresontario: oh yes totally, this is intended to be an exception...but I agree we should explicitly document why. The disabled is mainly about preventing navigation while the app is in a transient/unsafe state (for example while saving, resolving route guards, or when the current step shouldn’t allow leaving yet), and in `href` mode it also lets us apply proper aria-disabled/focus behaviour. I'll add a note regarding this.

packages/ontario-design-system-component-library/src/components/ontario-back-button/test/ontario-back-button.spec.tsx
erincarrieresontario: Yep, agreed.I removed them so unit tests only cover behaviour this component actually controls, and we keep real keyboard activation coverage in the Playwright/e2e suite.

packages/ontario-design-system-component-library/src/components/ontario-back-button/readme.md
```diff
@@ -0,0 +1,373 @@
...
+
+The back button is rendered as a tertiary-styled button (or link in href mode) with:
+
+- Inline flex layout with centered icon and text
+- Leading left-chevron icon (decorative)
+- Localized label ("Back" / "Retour")
+- Mobile-responsive width (full width behavior on small screens)
+- Consistent spacing and icon sizing tuned for the design system
+
+## Properties
```
smorrisods: This manually-written `## Properties` table sits *above* `<!-- Auto Generated Below -->`. I checked `ontario-back-to-top`'s readme for comparison — Stencil generates a real Properties table (plus Events/Slots and the `Built with StencilJS` footer) for the same kind of prop types, it's just that this readme's auto-generated section below the marker is empty (never run). Rather than hand-maintaining a stand-in, let's run the docs generator so this is the single source of truth — this table's already drifted: it lists `backMode`'s default as `"history"`, but the prop is `backMode?: BackMode` with no default; `history` is only the runtime-inferred fallback.

packages/ontario-design-system-component-library/src/components/ontario-back-button/readme.md
```diff
@@ -0,0 +1,373 @@
...
+<OntarioBackButton onBackClick={(event) => console.log('Clicked', event)} />
+```
+
+**Angular:**
+
+```html
+<ontario-back-button (backClick)="onBackClick($event)"></ontario-back-button>
+```
+
+<!-- Auto Generated Below -->
```
smorrisods, line 380: Nothing was generated below this marker — looks like `stencil build --docs` was never run for this component, so the Properties table above had to be hand-written as a stand-in. Let's run the docs build so this section gets populated automatically (Properties, Events, and the `Built with StencilJS` footer, same as the other components), then drop the manual table above.

smorrisods (changes requested):
This is in really solid shape — thorough JSDocs, a well-organized readme, and a strong test suite covering all three modes plus disabled/a11y behaviour. Nice cleanup on the search-box snapshot test too, even if slightly out of scope here.

One thing to fix before merging: the `## Properties` table in the readme is hand-written above `<!-- Auto Generated Below -->`, and nothing was ever generated below it — looks like the docs build step wasn't run (this component may have been scaffolded with AI assistance, which would explain it). The hand-written table's already drifted from the code too (`backMode` default listed as `"history"`, but it's actually runtime-inferred, not a component default). Let's run the docs generator and drop the manual table so there's one source of truth. Also carrying forward one still-open question on the `<a>`-count test assertion.

---
🤖 Beep boop — Ghostwritten by CoPilot — Human judgment (and sarcasm) still pending.

smorrisods:
E2E tests are also failing 😞

packages/ontario-design-system-component-library/src/components/ontario-back-button/test/ontario-back-button.e2e.ts
erincarrieresontario: yeah that check is there to make sure the "history mode" test is actually testing a button and not accidentally passing because of a link. Since this component can render either a <button> or an <a> depending on the mode, just checking if there a button with text "Back" isn't enough on its own

erincarrieresontario:
@smorrisods resolved all of the noted issues :)

packages/ontario-design-system-component-library/src/components/ontario-back-button/readme.md
```diff
@@ -0,0 +1,403 @@
...
+- Temporarily disabling the back control can prevent accidental navigation and data loss
+- In `href` mode, the component applies `aria-disabled` and removes tab focus to match disabled link semantics
+
+Use this sparingly. Prefer clear messaging whenever possible, and only disable when navigation must be temporarily blocked for correctness/safety.
+
+## Accessibility
+
+The back button uses native `<button>` or `<a>` semantics, making it keyboard and screen-reader compatible by default:
+
+- **Keyboard support:** Activates with Enter or Space keys
```
muntalee, line 42: This line says the control activates with Enter or Space, but the href render path appears to rely on click handling only (see `ontario-back-button.tsx:223`). 
\
Could we either implement explicit Space-key activation for the anchor path, or update this docs line to describe keyboard behaviour per mode?

packages/ontario-design-system-component-library/src/components/ontario-back-button/test/ontario-back-button.e2e.ts
```diff
@@ -0,0 +1,148 @@
...
+		await page.setContent('<input id="first" /><ontario-back-button id="back-btn"></ontario-back-button>');
+		await page.waitForChanges();
+
+		await page.locator('#first').focus();
+		await expect(page.locator('#first')).toBeFocused();
+
+		await page.keyboard.press('Tab');
+		await expect(page.locator('#back-btn').locator('button')).toBeFocused();
+	});
+});
```
muntalee, line 195: This block verifies keyboard focus, and href disabled semantics are covered above. To align with the accessibility guidance (and the docs), could we also add e2e assertions for Enter and Space activation behaviour in href mode?

packages/ontario-design-system-component-library/src/components/ontario-back-button/ontario-back-button.tsx
```diff
@@ -0,0 +1,233 @@
...
+	 * during guarded/transient states (for example, save-in-progress or step-transition lock).
+	 * In those cases, `disabled` prevents accidental navigation while preserving a consistent UI shell.
+	 */
+	@Prop() disabled?: boolean = false;
+
+	/**
+	 * Emitted when the user activates the back control.
+	 * Emitted before navigation when applicable.
+	 */
+	@Event() backClick!: EventEmitter<MouseEvent | KeyboardEvent>;
```
muntalee, line 75: This event is typed as MouseEvent or KeyboardEvent. I can see click-driven emission, but I do not see explicit keyboard listener wiring in this component. Could we either narrow the event detail type to MouseEvent, or add keyboard event handling so KeyboardEvent can actually be emitted?

muntalee (changes requested):
Hellooo I have some reviews mainly involved with keyboard interactions.

packages/ontario-design-system-component-library/src/components/ontario-back-button/readme.md
erincarrieresontario: Good call @muntalee - added here: https://github.com/ongov/ontario-design-system/pull/271/changes/551db90cfb4feaf3586cbb79bd9a07568c28e9ea

packages/ontario-design-system-component-library/src/components/ontario-back-button/test/ontario-back-button.e2e.ts
erincarrieresontario: Done - added here: https://github.com/ongov/ontario-design-system/pull/271/changes/de8e0d42a53825af28847f76fe1f4959a025a3ed

packages/ontario-design-system-component-library/src/components/ontario-back-button/ontario-back-button.tsx
erincarrieresontario: Actually already covered, I added a keydown handler on the anchor for the Space key (since native links don't activate on Space like buttons do), and it emits the real KeyboardEvent through backClick. So both types are genuinely used, no change needed.

packages/ontario-design-system-component-library/src/components/ontario-back-button/test/ontario-back-button.e2e.ts
smorrisods: Can we add a comment here to call this out - at this point the test has already checked for a button so if this is a special edge case we should note it where it is with a comment or make it its own test.

packages/ontario-design-system-component-library/src/components/ontario-back-button/ontario-back-button.tsx
smorrisods: Can we add to the comment that it should be used to disable the button based on user input, that should be validated via a function to check if going back is valid or not rather than locking a user out in normal usage.

I'm thinking of what an AI agent might scheme up to do with the back button given we're introducing a concept we've worked hard to not introduce and instead rely on validation rather than properties.

Which gives me another idea - maybe we farm this out to a function rather than an explicit property.  If the function is set then it should be used to validate the back state? Thoughts?

packages/ontario-design-system-component-library/src/components/ontario-back-button/ontario-back-button.tsx
```diff
@@ -0,0 +1,251 @@
...
+	private get resolvedBackMode(): BackMode {
+		if (this.backMode) {
+			return this.backMode;
+		}
+
+		if (this.href) {
+			return 'href';
+		}
+
+		return 'history';
```
smorrisods: This could be setup as a single line.

```suggestion
		return (this.backMode || this.href ? 'href' : 'history') ?? 'history';
```

packages/ontario-design-system-component-library/src/components/ontario-back-button/ontario-back-button.tsx
```diff
@@ -0,0 +1,251 @@
...
+	/**
+	 * Handles user activation for both button and link render paths.
+	 *
+	 * Emits `backClick` first, then applies mode-specific behaviour:
+	 * - `event`: emit only
+	 * - `history`: call `window.history.back()`
+	 * - `href`: allow native link navigation when href exists
+	 */
+	private handleActivation(event: MouseEvent | KeyboardEvent) {
+		if (this.disabled) {
```
smorrisods, line 162: eg. call out here for an async function to validate the button state.

packages/ontario-design-system-component-library/src/components/ontario-back-button/ontario-back-button.tsx
```diff
@@ -0,0 +1,251 @@
...
+			if (isClientSideRendering()) {
+				window.history.back();
+			}
+			return;
+		}
+
+		if (mode === 'href' && !this.href) {
+			event.preventDefault();
+		}
+	}
```
smorrisods, line 179: Every single mode is calling `event.preventDefault();` we can move this to the top of the method.

packages/ontario-design-system-component-library/src/components/ontario-back-button/ontario-back-button.tsx
smorrisods: I guess this is technically not a form control, so that should be called out in the JSDocs somewhere. To alert the humans and the robots.

packages/ontario-design-system-component-library/src/components/ontario-back-button/readme.md
```diff
@@ -0,0 +1,403 @@
...
+
+Use a back button to provide a clear, consistent way for users to return to a previous step.
+
+## When to use this component
+
+- User is progressing through a linear or mostly linear flow
+- User benefit comes from returning to the previous step in-context
+- You need a consistent, accessible back navigation pattern
+
+## When not to use this component
```
smorrisods, line 15: We should mention this is not a form component.

packages/ontario-design-system-component-library/src/components/ontario-back-button/readme.md
```diff
@@ -0,0 +1,403 @@
...
+
+- The page has multiple entry points and previous-page expectations are ambiguous
+- Breadcrumbs are already present (do not use both together)
+- The back action is unclear or would confuse users about where they're going
+
+## Usage guidance
+
+Please refer to the [Ontario Design System](https://designsystem.ontario.ca/components/detail/back-button.html) for current documentation guidance.
+
+## Disabled state rationale
```
smorrisods, line 26: Maybe disabled should just not render the component? show a show and hide feature vs. a disabled feature?

packages/ontario-design-system-component-library/src/components/ontario-back-button/readme.md
```diff
@@ -0,0 +1,403 @@
...
+
+## Component architecture
+
+This component is **intentionally navigation-strategy agnostic**. The back button itself is presentational—it does not encode routing logic. Instead, you choose the mode that fits your application's navigation approach:
+
+| Mode                  | Use When                                                            | Renders As | Behavior                                                   |
+| --------------------- | ------------------------------------------------------------------- | ---------- | ---------------------------------------------------------- |
+| **history** (default) | Users follow a linear flow and browser history is reliable          | `<button>` | Calls `window.history.back()`                              |
+| **href**              | You need deterministic navigation to a known previous step          | `<a>` link | Navigates to the specified URL                             |
+| **event**             | Your framework (React Router, Angular Router, etc.) manages routing | `<button>` | Emits `backClick` event only; your app controls navigation |
```
smorrisods: I suggest using back-ticks for the Mode since these are the actual code modes.

```suggestion
| `history` (default) | Users follow a linear flow and browser history is reliable          | `<button>` | Calls `window.history.back()`                              |
| `href`              | You need deterministic navigation to a known previous step          | `<a>` link | Navigates to the specified URL                             |
| `event`             | Your framework (React Router, Angular Router, etc.) manages routing | `<button>` | Emits `backClick` event only; your app controls navigation |
```

packages/ontario-design-system-component-library/src/components/ontario-back-button/readme.md
```diff
@@ -0,0 +1,403 @@
...
+```mdx-code-block
+<Tabs
+defaultValue="html"
+values={[
+{label: 'HTML', value: 'html'},
+{label: 'React', value: 'react'},
+{label: 'Angular', value: 'angular'},
+]}
+groupId="framework"
+queryString="framework">
```
smorrisods: I think this can be indented? 🤔

packages/ontario-design-system-component-library/src/components/ontario-back-button/readme.md
```diff
@@ -0,0 +1,403 @@
...
+
+```tsx
+<OntarioBackButton onBackClick={(event) => console.log('Clicked', event)} />
+```
+
+**Angular:**
+
+```html
+<ontario-back-button (backClick)="onBackClick($event)"></ontario-back-button>
+```
```
smorrisods: This would benefit from the Tabs approach of the other examples to remain consistent.

smorrisods (changes requested):
I've done my final detailed pass and noticed a few items worth commenting on.  One concern I have is the disabled state - I understand where it is coming from but it also leads to the similar challenges we have with disabling form elements stemming from the question, "why is the back button disabled?"  I'm now wondering if it is worth change it to a hidden property where we just don't show it at all if a user can't navigate back.  Thoughts?

muntalee (commented):
Hey lovely changes! Are you planning on adding any VRT in this PR? Or would you like to do it in a separate PR.

erincarrieresontario:
> I've done my final detailed pass and noticed a few items worth commenting on. One concern I have is the disabled state - I understand where it is coming from but it also leads to the similar challenges we have with disabling form elements stemming from the question, "why is the back button disabled?" I'm now wondering if it is worth change it to a hidden property where we just don't show it at all if a user can't navigate back. Thoughts?

yeah this is a good point...and it lines up with why we avoid disabled on other components too, like dropdown-list and textarea. One thing though, right now disabled is meant for temporary stuff, like mid-save or a locked step, not necessarilly "there's nowhere to go back to." Those feel like different problems to me i think, like if if there's genuinely no back destination, hiding makes total sense. But if back is just temporarily unsafe to click, hiding it might actually be more confusing since it'd cause layout shift and look broken? So maybe we keep both? Add a hidden prop for the "no destination" case, and keep disabled for the temporary lock case, just tighten up the docs so it's clear that's all disabled is for. Let me know if that sounds good to you @smorrisods

packages/ontario-design-system-component-library/src/components/ontario-back-button/ontario-back-button.tsx
erincarrieresontario: I can't seem to respond to the above note about the disabled prop, so I'm adding it here: Yeah, this actually goes against what we said earlier about keeping disabled for the temporary lock case. I think the canGoBack idea is better though, it makes you actually justify why back isn't available instead of just flipping a switch. So maybe we ditch that other prop entirely and use canGoBack for both situations since the app can just return false either way. Simpler, and it avoids the whole problem instead of just moving it somewhere else.

packages/ontario-design-system-component-library/src/components/ontario-back-button/ontario-back-button.tsx
erincarrieresontario: Done!

packages/ontario-design-system-component-library/src/components/ontario-back-button/readme.md
erincarrieresontario: Done!

packages/ontario-design-system-component-library/src/components/ontario-back-button/readme.md
erincarrieresontario: Done!

packages/ontario-design-system-component-library/src/components/ontario-back-button/readme.md
erincarrieresontario: Done !

packages/ontario-design-system-component-library/src/components/ontario-back-button/readme.md
erincarrieresontario: Done!

erincarrieresontario:
Hey @smorrisods , I wrote some feedback regarding the disabled prop, let me know what you think about my thoughts.

erincarrieresontario:
> Hey lovely changes! Are you planning on adding any VRT in this PR? Or would you like to do it in a separate PR.

Good call @muntalee , I've added VRT Tests :)

packages/ontario-design-system-component-library/src/components/ontario-back-button/test/ontario-back-button.vrt.ts
```diff
@@ -0,0 +1,82 @@
...
+ * shadow nodes, with animations disabled and the caret hidden for determinism.
+ */
+test.describe('ontario-back-button - default states', () => {
+	test('button mode - default state', async ({ page }) => {
+		await page.setContent(`<ontario-back-button></ontario-back-button>`);
+		await page.waitForChanges();
+
+		const host = page.locator('ontario-back-button');
+		await expect(host).toHaveClass('hydrated');
+		await expect(host).toHaveScreenshot('ontarioBackButton-default.png', {
```
smorrisods, line 22: Let's use the automatic naming scheme for these tests so we don't have to maintain the custom names.

muntalee (approved):
LGTM 🐐

smorrisods (changes requested):
I forgot to submit this from my review yesterday, sorry.  This will make these consistent with our other new VRTs.

smorrisods (approved):
LGTM 🚀

PR #277, opened 2026-06-22
DS-2710: Add Stencil visual regression testing job to CI

smorrisods:
## Summary

Adds a Visual Regression Testing (VRT) job for the Stencil component library, bringing the same model the demo apps already use (`*.vrt.ts` + Playwright `toHaveScreenshot`) to the component library as a companion to the existing Stencil E2E tests.

## Motivation

The component library had no automated way to catch unintended visual changes. The apps pipeline already runs VRT against committed baselines; this change closes that gap for the core web components so visual regressions are caught in CI rather than in review.

## Changes

- **Playwright config** (`packages/ontario-design-system-component-library/playwright.config.ts`)
  - Broaden `testMatch` to discover `*.vrt.ts` files — `@stencil/playwright`'s `createConfig()` only matches `*.e2e.ts` by default, so VRT files were otherwise invisible.
  - Add a `snapshotPathTemplate` mirroring the app-nextjs VRT layout and pin an explicit `chromium` project so the snapshot `{projectName}` token is stable.
  - Raise the dev `webServer.timeout` to 5 minutes; a cold Stencil dev build can take well over 2 minutes before it starts serving.
- **Suite split** (component library + root `package.json`)
  - Scope `test:e2e` to `*.e2e.ts` and add `test:vrt` / `test:update-snapshots` (with `retries: 0`) plus their `:docker` variants.
  - Add scoped `test:vrt:stencil` / `test:vrt:apps` helpers to the root scripts.
- **Bootstrap coverage** (`ontario-button.vrt.ts` + baselines)
  - Adds `ontario-button` VRT covering primary/secondary/tertiary default and focus states, with animations disabled and the caret hidden for determinism, plus the committed chromium baselines so the new job starts green.
- **Docker + CI**
  - Add a `stencil-vrt-runner` compose service using the shared Playwright runner anchor so contributors can refresh baselines in the same image CI uses.
  - Add a `stencil-vrt-test` job to the reusable test workflow that runs against the prebuilt artifacts and uploads/publishes the results, JUnit report and Playwright HTML report.
- **Docs**
  - Document the new commands and Docker runner, and correct the contributing guidelines testing section (E2E runs on Playwright, not Jest).

## Notes

- The suites are split at invocation time with a simple Playwright filter (`e2e.ts` / `vrt.ts`). These are Playwright regexes, not shell globs — quoted glob patterns like `**/*.vrt.ts` are interpreted as regular expressions and fail.
- Baselines are committed as chromium PNGs. Refresh them with `pnpm --filter @ongov/ontario-design-system-component-library run test:update-snapshots` (prefer the `:docker` variant to stay consistent with CI) whenever an intentional visual change is made.

## Testing

- `pnpm run test:vrt:stencil` — 6 passed against the committed baselines.


muntalee (approved):
LGTM 🐐

PR #285, opened 2026-06-23
Restore `ontario-summary-list-item` styles after clean Stencil build

muntalee:
The component rendered `ontario-summary-list-item__*` class names, but some stylesheet selectors were using `ontario-summary-list__*`, so the styles did not apply after the Stencil build.

- corrected the summary list item SCSS selectors to match the component's rendered BEM class names
- restored row and button-container styling for `ontario-summary-list-item`


smorrisods (approved):
LGTM 🤘🏻

PR #290, opened 2026-06-28
fix(deps): update dependency @uirouter/angular to v21

smorrisods:
This PR contains the following updates:

| Package | Change | [Age](https://docs.renovatebot.com/merge-confidence/) | [Confidence](https://docs.renovatebot.com/merge-confidence/) |
|---|---|---|---|
| [@uirouter/angular](https://ui-router.github.io/angular) ([source](https://redirect.github.com/ui-router/angular)) | [`^11.1.0` → `^21.0.0`](https://renovatebot.com/diffs/npm/@uirouter%2fangular/11.1.0/21.0.0) | age | confidence |

---

### Release Notes

<details>
<summary>ui-router/angular (@&#8203;uirouter/angular)</summary>

### [`v21.0.0`](https://redirect.github.com/ui-router/angular/blob/HEAD/CHANGELOG.md#2100-2026-01-03)

[Compare Source](https://redirect.github.com/ui-router/angular/compare/17.0.1...21.0.0)

[Compare `@uirouter/angular` versions 17.0.1 and 21.0.0](https://redirect.github.com/ui-router/angular/compare/17.0.1...21.0.0)

##### Bug Fixes

- support zoneless ([#&#8203;1012](https://redirect.github.com/ui-router/angular/issues/1012)) ([6fa2d1e](https://redirect.github.com/ui-router/angular/commit/6fa2d1e))

##### Features

- Support Angular v21 ([4c66883](https://redirect.github.com/ui-router/angular/commit/4c66883))

#### 17.0.1 (2025-12-31)

[Compare `@uirouter/angular` versions 17.0.0 and 17.0.1](https://redirect.github.com/ui-router/angular/compare/17.0.0...17.0.1)

***

##### Updated `@uirouter/core` from 6.1.1 to 6.1.2

Changelog for `@uirouter/core`:

[Compare `@uirouter/core` versions 6.1.1 and 6.1.2](https://redirect.github.com/ui-router/core/compare/6.1.1...6.1.2)

##### Bug Fixes

- **urlMatcher:** Fix potential ReDoS, thanks to [www.HeroDevs.com](http://www.HeroDevs.com) ([0db52a1](https://redirect.github.com/ui-router/core/commit/0db52a1))

### [`v17.0.1`](https://redirect.github.com/ui-router/angular/blob/HEAD/CHANGELOG.md#2100-2026-01-03)

[Compare Source](https://redirect.github.com/ui-router/angular/compare/17.0.0...17.0.1)

[...]

PR #289, opened 2026-06-28
fix(deps): update dependency @ngx-translate/http-loader to v18

smorrisods:
This PR contains the following updates:

| Package | Change | [Age](https://docs.renovatebot.com/merge-confidence/) | [Confidence](https://docs.renovatebot.com/merge-confidence/) |
|---|---|---|---|
| @&#8203;ngx-translate/http-loader | [`^17.0.0` → `^18.0.0`](https://renovatebot.com/diffs/npm/@ngx-translate%2fhttp-loader/17.0.0/18.0.0) | age | confidence |

---

### Configuration

📅 **Schedule**: (in timezone UTC)

- Branch creation
  - "before 3am on Sunday"
- Automerge
  - At any time (no schedule defined)

🚦 **Automerge**: Disabled by config. Please merge this manually once you are satisfied.

♻ **Rebasing**: Whenever PR is behind base branch, or you tick the rebase/retry checkbox.

🔕 **Ignore**: Close this PR and you won't be reminded about this update again.

---

 - [ ] If you want to rebase/retry this PR, check this box

---

This PR has been generated by [Mend Renovate](https://redirect.github.com/renovatebot/renovate).

PR #288, opened 2026-06-28
fix(deps): update dependency @ngx-translate/core to v18

smorrisods:
This PR contains the following updates:

| Package | Change | [Age](https://docs.renovatebot.com/merge-confidence/) | [Confidence](https://docs.renovatebot.com/merge-confidence/) |
|---|---|---|---|
| [@ngx-translate/core](https://redirect.github.com/ngx-translate/core) | [`^17.0.0` → `^18.0.0`](https://renovatebot.com/diffs/npm/@ngx-translate%2fcore/17.0.0/18.0.0) | age | confidence |

---

### Release Notes

<details>
<summary>ngx-translate/core (@&#8203;ngx-translate/core)</summary>

### [`v18.0.0`](https://redirect.github.com/ngx-translate/core/releases/tag/v18.0.0)

[Compare Source](https://redirect.github.com/ngx-translate/core/compare/v17.0.0...v18.0.0)

`@ngx-translate/core@18.0.0` and `@ngx-translate/http-loader@18.0.0` are now stable on npm (`latest`).

v18 is the **major API-cleanup release**: standalone providers everywhere, signal-based internals, the `default` → `fallback` rename finalized, and the deprecated surface removed. It also resolves three long-standing semantic gotchas and baselines the workspace on Angular 22 / TypeScript 6.0.

#### Highlights

- **Standalone-first setup.** `TranslateModule` is gone — use `provideTranslateService()` / `provideChildTranslateService()` (works in standalone *and* NgModule apps).
- **New `*translateBlock="let t"` structural directive** for cleaner template translations.
- **New `isLoading: Signal<boolean>`** — true while any `use()` load is in flight; read it for language-switching spinners.
- **Isolated subtrees are now truly isolated** — own events, lookups, and language state.
- **Angular 18–22 supported** (16 and 17 dropped); built and tested against Angular 18/19/20/21/22.

#### Breaking changes — required edits

**`TranslateModule` removed.** Replace `TranslateModule.forRoot({...})` / `forChild({...})` with `provideTranslateService({...})` / `provideChildTranslateService({...})`. The pipe, directive, and service are standalone.

**Deprecated `default*` symbols removed — rename:**

| Removed                  | Replacement               |
| ------------------------ | ------------------------- |
| `defaultLang`            | `fallbackLang`            |
| `setDefaultLang()`       | `setFallbackLang()`       |
| `getDefaultLang()`       | `getFallbackLang()`       |
| `onDefaultLangChange`    | `onFallbackLangChange`    |
[...]

PR #292, opened 2026-07-10
Add build-apps CI job to catch app-breaking dependency bumps

smorrisods:
## Summary

Adds a `build-apps` job to CI so that `app-angular` and `app-react` are actually built on every PR, not just `app-nextjs`. Right now, dependency bumps (Renovate or otherwise) can merge with a fully green CI pipeline even if they break the Angular or React demo apps, because `build.yml` only runs `pnpm run build-libs`, and `test.yml`'s app-scoped jobs silently skip `app-angular`/`app-react` since neither package defines a `test:e2e`/`test:vrt` script for lerna to run. `app-nextjs` only gets exercised indirectly, as a side effect of its own Playwright `webServer` script (`scripts/start-server.sh`), which builds and starts it before running tests against it.

This was discovered while reviewing #288, #289, and #290 (the `@ngx-translate/core`, `@ngx-translate/http-loader`, and `@uirouter/angular` Renovate PRs): `app-angular` currently fails to build on `develop` due to a duplicate `@angular/core` version resolution (`22.0.0` vs `22.0.5`), and `@ngx-translate/core` v18 removes `TranslateModule` entirely, breaking the app's translation setup. Neither of these would have surfaced in CI without manually building the app locally.

## Changes

- Renamed the existing `build` job in `build.yml` to `build-libs` for clarity, since there are now two build jobs.
- Added a `build-apps` job (`needs: ['build-libs']`) that downloads the `build-libs` artifact, installs dependencies, and runs `pnpm run build-apps`, which covers `app-angular`, `app-react`, and `app-nextjs` in one shot (via the existing `lerna run build --scope 'app-*'` script).
- The `build-apps` job uploads its own artifact (`app-build-artifacts`) containing the built output of each app (`app-angular/dist`, `app-react/build`, `app-nextjs/.next`).
- Threaded the new `app-build-artifacts-name` input/output through `main.yml` and `test.yml` so downstream jobs can consume it.
- `apps-e2e-test` and `apps-vrt-test` in `test.yml` now download the `app-build-artifacts` artifact and set `SKIP_APP_BUILD=true` when invoking Playwright, so `app-nextjs` isn't rebuilt a second and third time in the pipeline (it was previously being built once per Playwright job via `start-server.sh`).
- Updated `scripts/start-server.sh` to skip its own `pnpm run build` step when `SKIP_APP_BUILD=true` is set, since the app was already built upstream.

## Motivation

The Angular and React demo apps are not currently covered by CI at all from a build-correctness standpoint, which means dependency bumps that compile fine for the shared libraries can still merge and silently break either app. This is a structural gap that should be closed before we keep merging dependency-bump PRs against those apps. This PR only wires up the build step; it does not add unit or E2E test coverage for `app-angular`/`app-react` (there is no test runner configured for `app-react` yet, and `app-angular`'s Karma/Jasmine unit tests aren't wired into CI either) — that's tracked as follow-up work.

## Testing

This has not yet been exercised against an actual CI run; the workflow YAML has been validated for syntax (`yaml.safe_load`) and formatting (`prettier --check`), and the underlying `pnpm run build-apps` command has been run locally against this branch to confirm it produces the expected output directories. Verifying the full pipeline (artifact upload/download, the `SKIP_APP_BUILD` short-circuit, etc.) will happen once this PR runs in GitHub Actions.

## Related

Surfaced during review of #288, #289, and #290.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,6 +9,9 @@ on:
       docs-artifacts-name:
         required: true
         type: string
+      app-build-artifacts-name:
+        required: true
+        type: string
     outputs: # https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs
       build-artifacts-name:
         description: 'Name of the artifacts archive of the built libraries'
@@ -16,9 +19,12 @@ on:
       docs-artifacts-name:
         description: 'Name of the artifacts archive of the built docs'
         value: ${{ inputs.docs-artifacts-name }}
+      app-build-artifacts-name:
+        description: 'Name of the artifacts archive of the built apps'
+        value: ${{ inputs.app-build-artifacts-name }}
 
 jobs:
-  build:
+  build-libs:
     runs-on: ubuntu-latest
     steps:
       # Checkout the repo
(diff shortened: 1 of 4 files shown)


smorrisods:
Closed automatically by GitHub when the head branch was renamed to `scott/ci/build-apps-job` (to follow the repo's branch naming convention). Reopened as #293.

PR #299, opened 2026-07-13
DS-2690: Author non-colour primitive tokens

smorrisods:
## Summary
- Lands the space, typography, radius, border, elevation, and motion primitive tokens from the reviewed prototype JSON.
- Authors `z-index.json` and `breakpoint.json` from the existing legacy `ontarioZIndex` / `ontarioBreakpoints` values (`tokens/z-index/z-index.js`, `tokens/breakpoints/breakpoints.js`) so the stacking scale and responsive breakpoints carry forward unchanged rather than being invented.
- `breakpoint.json` values are authored directly in `em` (already em-native in the legacy source), so the `pxToEm` transform doesn't need to touch them.
- Every token carries an explicit `type` (`borderWidth`, `borderRadius`, `boxShadow`, `spacing`, `sizing`, `fontFamilies`, `fontSizes`, `fontWeights`, `lineHeights`, `letterSpacing`, `duration`, `cubicBezier`, `other`), matching the vocabulary already used for the colour primitives and the GC Design System's token export convention (github.com/cds-snc/gcds-tokens). This is the single source of truth downstream export tooling (DS-2692) reads from — none of it infers a type from a token's path or value.
- Extends `test/tokens.test.mjs` to assert all eight non-colour category files ship.

## Why
DS-2690 completes the remaining ~84 non-colour primitives of the Core (primitive) layer, alongside the colour primitives landed in DS-2687/DS-2688. Independent of colour — able to land in parallel.

Two related legacy categories are **deliberately left out** and tracked as design-decision issues instead of being ported directly, since neither the Design Token Review report nor the DS-2685 plan ever reviewed them as tokens:
- Container/max-width scale (`widthNarrower`/`widthNarrow`/`widthStandard`/`widthMax`) → **DS-2730**
- 12-column grid config (`grid-columns.js`) → **DS-2731**

Both are children of DS-2330 (the design-owned primitive audit epic); a follow-up Develop issue will implement whatever's decided.

## Validation
- `pnpm run lint:tokens`: 22 files checked, 0 warnings, 0 errors.
- `pnpm test`: 8 tests passing (transforms + primitive file/linter integrity).
- `node build.mjs`: `dist/` byte-identical to the pre-change baseline (zero behaviour change).

## Base branch note
This PR targets `scott/task/DS-2687-primitive-sources-tooling` (PR #284, not yet merged) rather than the epic branch directly, since the `tokens/primitives/` structure and tooling this depends on only exist there so far. Rebase onto the epic branch once #284 merges.


packages/ontario-design-system-design-tokens/test/tokens.test.ts
erincarrieresontario, line 1: hmm it looks like these tests only verify filenames, so malformed or incomplete token values could still pass. Could this test load each file and assert the expected token paths and value shapes?

packages/ontario-design-system-design-tokens/tokens/primitives/breakpoint.json
erincarrieresontario, line 1: this authors values directly as em, while the primitive transform layer expects source breakpoint values in px before converting them to em. Should these be authored in px for consistency, or should the transform/input policy be explicitly updated?

erincarrieresontario (approved):
Looks good, just a few comments @smorrisods :)

muntalee (changes requested):
Small note: the description says these tokens all have an explicit type field, but I don't see one in any of the new JSON files — looks like they're relying on the auto-detection in #353 instead. Not a blocker, just want to flag it in case the description should be updated.

PR #300, opened 2026-07-13
DS-2721, DS-2732: Add pre-flight pnpm/drift checks

smorrisods:
## Summary

Fixes two related CI reliability bugs that both surfaced during the `v9.0.0-alpha.1` release: a pnpm version mismatch between this monorepo and the docs repo broke `sync-docs`, and Stencil-generated tracked file drift wasn't caught until the release pipeline's dirty-tree check, which was silently masked by a `git reset --hard HEAD`.

### DS-2732: Stencil-generated tracked files drift undetected until release
- Adds a drift-check step to `build.yml`, run immediately after `pnpm run build-libs`, that fails the build if any of the tracked Stencil-generated files (`components.d.ts` in both the component library and HTML generator packages, the Angular `stencil-generated/*` wrappers, and the 36 component `readme.md` files) differ from what's committed. Fails fast with the list of drifted files and the fix (`pnpm build-libs`, commit, push).
- Removes the `git reset --hard HEAD` safety net in `publish-packages.yml` entirely (previously used to silently force a clean tree before `semantic-release-lerna`'s dirty-tree check). It's no longer needed since drift is now caught earlier, and it was a liability that hid real problems.

### DS-2721: pnpm version mismatch between monorepo and docs repo breaks sync-docs
- Adds a new reusable `pre-flight.yml` workflow, wired into `main.yml` between `build`/`test` and `publish-packages`, that compares the `packageManager` field in this repo against the docs repo's `package.json` (via `gh api`) and fails fast on mismatch, before any packages are published.
- Restores `--frozen-lockfile` for the initial dependency install step in `sync-docs-to-docusaurus.yml`. The subsequent `pnpm add` step that bumps the component-library version is a separate, intentional lockfile mutation and is unaffected by this flag.

## Out of scope (tracked separately / different repo)
- Aligning `design-system-developer-documentation-site`'s own `packageManager` field and CI (`pnpm/action-setup@v6`) — a follow-up change in that repo.
- Creating the missing `v9.0.0-alpha.1` docs PR.

## Test plan
- Verified all modified/new workflow YAML parses correctly.
- Will confirm in CI: `build` fails when generated files are stale, `pre-flight` fails on an intentionally mismatched pnpm version and blocks `publish-packages`, and `sync-docs-to-docusaurus` still succeeds with `--frozen-lockfile` on the initial install.

Jira: DS-2721, DS-2732


muntalee (approved):
LGTM 🐐

PR #301, opened 2026-07-13
DS-2733: Fix stale ontario-task taskStatus documentation

smorrisods:
## Summary

Fixes DS-2733: the `ontario-task` component's `taskStatus` documentation still described the old `TaskStatuses` enum semantics (PascalCase key names like `NotStarted`, `InProgress`) even though the underlying type was remodelled (in `c9083ac6`) to a `const` object whose valid values are camelCase runtime strings (`notStarted`, `inProgress`, etc.). This produced a JSDoc comment that directly contradicted the component's own generated `Type` column, and a hand-written SSR example that used an actually-invalid value.

## What changed

- **`ontario-task.tsx`**: the `taskStatus` `@Prop()` JSDoc no longer re-lists the accepted values — that's already fully expressed by the TS type and the Stencil-generated `Type` column in the readme, and restating it in prose is exactly what went stale here. The comment now only documents the fallback behaviour (falls back to the default on an invalid/missing value).
- **`ontario-task/readme.md`** (hand-written sections above `<!-- Auto Generated Below -->`):
  - SSR example: `task-status="InProgress"` → `task-status="inProgress"` (the old value would fail the component's own case-sensitive validator and silently fall back to the default at runtime).
  - Validation bullet: `'NotStarted'` → `'notStarted'`.
- **`ontario-task-list/readme.md`**: fixed a similar hand-written example using `data-task-status="Completed"` / `"Incomplete"` (the latter isn't even a valid status) → `"completed"` / `"notStarted"`.
- **`ontario-task.e2e.ts`**: fixed casing in stale **commented-out** test code while in the area (left commented, not re-enabled — re-enabling this suite is a separate decision).
- Regenerated `components.d.ts` / `readme.md` via `pnpm build-libs` to pick up the corrected JSDoc.

## Notes for reviewers

- This branch is stacked on `scott/ci/pre-flight-checks` (#300) since that PR's Stencil-drift-check work is what surfaced this bug (regenerating the previously-stale readme revealed the JSDoc/Type conflict). The diff here is scoped to just this fix; #300's changes are the CI/build.yml/pre-flight work.
- No behavioural/runtime code changed — this is documentation only.

## Testing

- `pnpm build-libs` succeeds; regenerated files verified.
- `vitest run src/components/ontario-task` — all 7 tests pass (ontario-task + ontario-task-list specs).

### DIFF
--- a/packages/ontario-design-system-component-library/src/components.d.ts
+++ b/packages/ontario-design-system-component-library/src/components.d.ts
@@ -3327,7 +3327,7 @@ export namespace Components {
 		 */
 		taskId: string;
 		/**
-		 * Defines the status of the task, with default set to 'NotStarted'.  Accepts values from `TaskStatus` const: `NotStarted`, `InProgress`, `Completed`, etc.
+		 * Defines the status of the task.  Falls back to the default status if not provided, or if an unrecognized value is passed.
 		 * @default TaskStatus.NotStarted
 		 */
 		taskStatus: TaskStatus;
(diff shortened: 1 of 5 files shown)


erincarrieresontario (approved):
LGTM 👍

muntalee (approved):
LGTM 🐐

PR #302, opened 2026-07-14
DS-2682: Implement autocomplete functionality in `ontario-search-box` component

muntalee:
## Summary
Adds complete autocomplete support to the `ontario-search-box` component with keyboard navigation, accessibility semantics, and flexible suggestion authoring modes (async suggestions, slotted `ontario-search-result-item`, and custom slotted HTML).

After the review comments, this was reworked to a segment-based highlight model so highlight behaviour is consistent across render paths.

## Merge Notice
This PR is dependent on DS-2703 (`ontario-search-result-item` component): https://github.com/ongov/ontario-design-system/pull/303

## Changes
- Enhanced ontario-search-box.tsx:
  - Props for autocomplete behaviour: `enableAutocomplete`, `getSuggestions`, `minChars`, `debounceMs`, `maxSuggestions`
  - Suggestion list state management (open/close/active/hover)
  - Keyboard navigation (Arrow keys, Enter, Escape, Tab)
  - ARIA combobox/listbox semantics (`aria-expanded`, `aria-controls`, `aria-activedescendant`, `aria-live`)
  - Selection handling for keyboard and mouse
  - Events: `autocompleteQueryUpdated`, `autocompleteSuggestionsUpdated`, `autocompleteSuggestionSelected`
- Introduced shared highlight model:
  - Added `Segment` type and `computeHighlightSegments` utility
  - Unified suggestion highlight behaviour (contiguous-first, fuzzy fallback)
- Updated `ontario-search-result-item`:
  - Added `segments` prop for highlight rendering
  - Kept legacy highlight fallbacks for compatibility
- Updated styling in `ontario-search-box.scss` to preserve visual behaviour and address review feedback
- Expanded test coverage:
  - Condition-based e2e checks for rendered autocomplete UI
  - Unit coverage for keyboard behaviour and highlight rendering paths

- Async suggestions via `getSuggestions(query)`
- Slotted `ontario-search-result-item` suggestions
- Slotted custom HTML suggestions

### DIFF
--- a/packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
+++ b/packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
@@ -15,7 +15,7 @@ import {
 	InputInteractionEvent,
 	InputInputEvent,
 } from '../../utils/events/event-handler.interface';
-import { computeHighlightSegments, Segment } from '../../utils/components/search-box-autocomplete';
+import { resolveSuggestionSegments, Segment } from '../../utils/components/search-box-autocomplete';
 
 import translations from '../../translations/global.i18n.json';
 
@@ -28,7 +28,9 @@ export interface AutocompleteSuggestion {
 	href?: string;
 	disabled?: boolean;
 	segments?: Segment[];
+	/** @deprecated Use `segments` instead. */
 	boldRanges?: Array<{ start: number; end: number }>;
+	/** @deprecated Use `segments` instead. */
 	highlightParts?: Array<{ text: string; isInputMatch: boolean }>;
 }
 
@@ -272,6 +274,8 @@ export class OntarioSearchBox {
 	private static readonly DEFAULT_MIN_CHARS = 1;
 	private static readonly DEFAULT_DEBOUNCE_MS = 150;
 	private static readonly DEFAULT_MAX_SUGGESTIONS = 8;
+	// Allows suggestion mousedown handlers to run before blur closes the list.
+	private static readonly BLUR_CLOSE_DELAY_MS = 120;
 	private static readonly RESULT_ITEM_TAG = 'ONTARIO-SEARCH-RESULT-ITEM';
 
 	/**
@@ -368,7 +372,7 @@ export class OntarioSearchBox {
 		}
 
 		if (eventType === EventType.Blur && this.enableAutocomplete) {
-			window.setTimeout(() => this.closeSuggestions(), 120);
+			window.setTimeout(() => this.closeSuggestions(), OntarioSearchBox.BLUR_CLOSE_DELAY_MS);
 		}
 
 		handleInputEvent(
@@ -388,6 +392,9 @@ export class OntarioSearchBox {
 		);
 	}
 
+	/**
+	 * Handles autocomplete query updates for slotted and async suggestion modes.
+	 */
 	private async handleAutocompleteInput(query: string) {
 		this.autocompleteQueryUpdated.emit({ query });
 
@@ -468,7 +475,7 @@ export class OntarioSearchBox {
 				id: `${this.getSuggestionListId()}-option-${index}`,
 				label: item,
 				value: item,
-				segments: this.resolveSuggestionSegments(item, query),
+				segments: resolveSuggestionSegments(item, query),
 			};
 		}
 
@@ -481,66 +488,10 @@ export class OntarioSearchBox {
 			description: item.description,
 			href: item.href,
 			disabled: !!item.disabled,
-			segments: this.resolveSuggestionSegments(label, query, item),
+			segments: resolveSuggestionSegments(label, query, item),
 			boldRanges: item.boldRanges,
 			highlightParts: item.highlightParts,
 		};
-	}
-
-	private resolveSuggestionSegments(label: string, query: string, suggestion?: AutocompleteSuggestion): Segment[] {
-		if (suggestion?.segments?.length) {
-			return suggestion.segments;
-		}
-
-		if (suggestion?.highlightParts?.length) {
-			return suggestion.highlightParts.map((part) => ({
-				text: part.text,
-				kind: part.isInputMatch ? 'match' : 'completion',
-			}));
-		}
-
-		if (suggestion?.boldRanges?.length) {
-			const ranges = suggestion.boldRanges
-				.map((range) => ({
-					start: Math.max(0, Math.min(range.start, label.length)),
-					end: Math.max(0, Math.min(range.end, label.length)),
-				}))
-				.filter((range) => range.end > range.start)
-				.sort((a, b) => a.start - b.start);
-
-			if (ranges.length) {
-				const mergedRanges: Array<{ start: number; end: number }> = [];
-				for (const range of ranges) {
-					const lastRange = mergedRanges[mergedRanges.length - 1];
-					if (!lastRange || range.start > lastRange.end) {
-						mergedRanges.push({ ...range });
-					} else {
-						lastRange.end = Math.max(lastRange.end, range.end);
-					}
-				}
-
-				const segments: Segment[] = [];
-				let cursor = 0;
-				for (const range of mergedRanges) {
-					if (cursor < range.start) {
-						segments.push({ text: label.slice(cursor, range.start), kind: 'match' });
-					}
-					segments.push({ text: label.slice(range.start, range.end), kind: 'completion' });
-					cursor = range.end;
-				}
-				if (cursor < label.length) {
-					segments.push({ text: label.slice(cursor), kind: 'match' });
-				}
-
-				return segments;
-			}
-		}
-
-		if (!(query || '').trim()) {
-			return [{ text: label, kind: 'completion' }];
-		}
-
-		return computeHighlightSegments(label, query) ?? [{ text: label, kind: 'completion' }];
 	}
 
 	/**
@@ -574,6 +525,9 @@ export class OntarioSearchBox {
 		return this.getAllSlotSuggestionElements(slot).filter((el) => !el.hasAttribute('hidden'));
 	}
 
+	/**
+	 * Updates slot state from the current assigned elements and reapplies option decoration.
+	 */
 	private updateSuggestionSlotState(slot?: HTMLSlotElement) {
 		const assignedOptions = this.getAllSlotSuggestionElements(slot);
 		this.hasSuggestionSlotContent = assignedOptions.length > 0;
@@ -589,22 +543,15 @@ export class OntarioSearchBox {
 		return this.getSuggestionValueFromOption(option);
 	}
 
-	private updateSlotSuggestionVisibility(query: string, assignedOptions = this.getAllSlotSuggestionElements()) {
-		const normalizedQuery = (query || '').trim();
-		const hasQuery = normalizedQuery.length > 0;
-
+	/**
+	 * Mirrors consumer-provided slot visibility into ARIA state without deciding relevance.
+	 */
+	private updateSlotSuggestionVisibility(_query: string, assignedOptions = this.getAllSlotSuggestionElements()) {
 		assignedOptions.forEach((option) => {
-			const optionLabel = this.getSlotOptionLabel(option);
-			const shouldShow = !hasQuery || computeHighlightSegments(optionLabel, normalizedQuery) !== null;
+			const isHidden = option.hasAttribute('hidden');
+			option.setAttribute('aria-hidden', String(isHidden));
 
-			if (shouldShow) {
-				option.removeAttribute('hidden');
-			} else {
-				option.setAttribute('hidden', '');
-			}
-			option.setAttribute('aria-hidden', String(!shouldShow));
-
-			if (!shouldShow) {
+			if (isHidden) {
 				option.removeAttribute('data-ontario-suggestion-index');
 				option.setAttribute('aria-selected', 'false');
 				option.classList.remove('ontario-search-autocomplete__slot-option--active');
@@ -630,6 +577,9 @@ export class OntarioSearchBox {
 		return option.children.length === 0 ? option : undefined;
 	}
 
+	/**
+	 * Decorates custom slotted markup that opts in via `data-ontario-search-highlight`.
+	 */
 	private decorateCustomSlotSuggestionOption(option: HTMLElement) {
 		const highlightTarget = this.getCustomSlotHighlightTarget(option);
 		if (!highlightTarget) {
@@ -642,7 +592,7 @@ export class OntarioSearchBox {
 			highlightTarget.dataset.ontarioSearchOriginalText = originalText;
 		}
 
-		const segments = this.resolveSuggestionSegments(originalText, this.value || '');
+		const segments = resolveSuggestionSegments(originalText, this.value || '');
 		const highlightNodes = segments.map((segment) => {
 			const segmentNode = document.createElement(segment.kind === 'completion' ? 'strong' : 'span');
 			segmentNode.className =
@@ -663,6 +613,9 @@ export class OntarioSearchBox {
 		});
 	}
 
+	/**
+	 * Applies accessibility attributes and active/hover styling hooks to slot options.
+	 */
 	private decorateSlotSuggestionOptions(assignedOptions = this.getSlotSuggestionElements()) {
 		assignedOptions.forEach((option, index) => {
 			if (!option.id) {
@@ -681,11 +634,11 @@ export class OntarioSearchBox {
 			option.classList.toggle('ontario-search-autocomplete__slot-option--active', isActive);
 			option.classList.toggle('ontario-search-autocomplete__slot-option--hovered', isHovered);
 
-			if (option.tagName === 'ONTARIO-SEARCH-RESULT-ITEM') {
+			if (option.tagName === OntarioSearchBox.RESULT_ITEM_TAG) {
 				(option as any).active = isActive;
 				const semanticLabel = (option as any).label || this.getSuggestionValueFromOption(option);
 				if (typeof semanticLabel === 'string' && semanticLabel.length) {
-					(option as any).segments = this.resolveSuggestionSegments(semanticLabel, this.value || '', option as any);
+					(option as any).segments = resolveSuggestionSegments(semanticLabel, this.value || '', option as any);
 				}
 			} else {
 				this.decorateCustomSlotSuggestionOption(option);
@@ -851,21 +804,15 @@ export class OntarioSearchBox {
 		this.suggestions = [];
 	}
 
+	/**
+	 * Selects a suggestion by index, updates the input value, and emits selection metadata.
+	 */
 	private selectSuggestionByIndex(index: number, source: 'keyboard' | 'mouse'): void {
 		if (index < 0 || this.isSuggestionDisabled(index)) {
 			return;
 		}
 
-		let selectedSuggestion:
-			| {
-					id?: string;
-					label: string;
-					value?: string;
-					description?: string;
-					href?: string;
-					disabled?: boolean;
-			  }
-			| undefined;
+		let selectedSuggestion: AutocompleteSuggestion | undefined;
 
 		if (this.hasSuggestionSlotContent) {
 			const option = this.getSlotSuggestionElements()[index];
@@ -942,6 +889,9 @@ export class OntarioSearchBox {
 		this.selectSuggestionByIndex(selectedIndex, 'mouse');
 	};
 
+	/**
+	 * Handles autocomplete keyboard navigation and keyboard-based suggestion selection.
+	 */
 	private handleInputKeyDown = (event: KeyboardEvent) => {
 		if (!this.enableAutocomplete) return;
 


packages/ontario-design-system-component-library/src/index.html
```diff
@@ -1612,6 +1695,38 @@ <h2 id="search-box">Search box</h2>
...
 						></ontario-search-box>
+
+						<h3 class="ontario-h4">Autocomplete with Ontario cities (async)</h3>
+						<ontario-search-box
+							id="ontario-search-autocomplete"
+							autocomplete
+							caption='{
+							"captionText": "Search Ontario cities",
+							"captionType": "default"
+						}'
```
smorrisods: The indentation is off here:


```suggestion
							caption='{
								"captionText": "Search Ontario cities",
								"captionType": "default"
							}'
```

packages/ontario-design-system-component-library/src/components/ontario-search-box/readme.md
```diff
@@ -123,6 +123,43 @@ The following example registers a simple function on `window` `load` that adds a
 </script>
 ```
 
+## Autocomplete examples
```
smorrisods, line 126: The examples here should also have React and Angular versions of them.

packages/ontario-design-system-component-library/src/components/ontario-search-box/readme.md
```diff
@@ -123,6 +123,43 @@ The following example registers a simple function on `window` `load` that adds a
...
+		const searchBox = document.getElementById('search-with-autocomplete');
+
+		searchBox.getSuggestions = async (query) => {
+			return cities.filter((city) => city.toLowerCase().includes((query || '').toLowerCase()));
+		};
+	});
+</script>
+```
+
+### Slotted semantic and custom HTML suggestions
```
smorrisods, line 224: This is cool! Good addition.  Perhaps we an add an example that has a static title in the list of suggestions that isn't selectable by the user, which is what is happening with the Ontario.ca search.

Example search results with internal sections

packages/ontario-design-system-component-library/src/components/ontario-search-box/readme.md
```diff
@@ -123,6 +123,43 @@ The following example registers a simple function on `window` `load` that adds a
...
+		<span data-ontario-search-highlight>Waterloo</span>
+		<span class="ontario-search-result-meta">Custom HTML option</span>
+	</div>
+</ontario-search-box>
+```
+
+Slot content takes precedence over `getSuggestions(query)` when both are supplied.
+For custom HTML suggestions, plain text-only options are highlighted automatically.
+If your custom option contains extra markup, wrap the text that should receive autocomplete bolding in an element with `data-ontario-search-highlight`.
+In slot mode, both semantic (`ontario-search-result-item`) and custom HTML options are filtered by the current query, and non-matching options are hidden.
```
smorrisods: These will render as one big paragraph since there are no line breaks in them.  Consider rewording them into two smaller paragraphs or making each one it's own paragraph.

Image

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -70,6 +74,47 @@ export class OntarioSearchBox {
 	 */
 	@Prop({ mutable: true }) value?: string;
 
+	/**
+	 * Enables autocomplete behaviour on the search input.
+	 */
+	@Prop() autocomplete?: boolean = false;
```
smorrisods: I recommend changing this to `autocompleteEnabled` since it is a boolean flag.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -70,6 +74,47 @@ export class OntarioSearchBox {
...
+			| {
+					id?: string;
+					label: string;
+					value?: string;
+					description?: string;
+					href?: string;
+					disabled?: boolean;
+					boldRanges?: Array<{ start: number; end: number }>;
+					highlightParts?: Array<{ text: string; isInputMatch: boolean }>;
+			  }
```
smorrisods: Let's define this as a named `type` rather than as an inline anonymous type.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -70,6 +74,47 @@ export class OntarioSearchBox {
...
+		)[]
+	>;
+
+	/**
+	 * Minimum number of characters required before suggestions are shown.
+	 */
+	@Prop() minChars?: number = 1;
+
+	/**
+	 * Debounce delay before `getSuggestions` is called.
```
smorrisods: This doc comment should mention it is in milliseconds.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -83,7 +128,7 @@ export class OntarioSearchBox {
 	 * >
 	 * </ontario-search-box>
 	 */
-	@Prop() caption: Caption | string;
+	@Prop() caption!: Caption | string;
```
smorrisods, line 142: This looks like this is mapped as required?

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -153,42 +198,86 @@ export class OntarioSearchBox {
...
 	 * 	</script>
 	 */
-	@Event() searchOnSubmit: EventEmitter<string>;
+	@Event() searchOnSubmit!: EventEmitter<string>;
 
 	/**
 	 * Emitted when a input  occurs when an input has been changed.
 	 */
-	@Event() inputOnInput: EventEmitter<InputInputEvent>;
+	@Event() inputOnInput!: EventEmitter<InputInputEvent>;
```
smorrisods, line 217: Does the `!` mark these as required?

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -153,42 +198,86 @@ export class OntarioSearchBox {
...
+			id?: string;
+			label: string;
+			value?: string;
+			description?: string;
+			href?: string;
+			disabled?: boolean;
+			boldRanges?: Array<{ start: number; end: number }>;
+			highlightParts?: Array<{ text: string; isInputMatch: boolean }>;
+		};
+		source: 'keyboard' | 'mouse';
```
smorrisods: This appears to be the same type as above.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -153,42 +198,86 @@ export class OntarioSearchBox {
...
+	private suggestions: {
+		id?: string;
+		label: string;
+		value?: string;
+		description?: string;
+		href?: string;
+		disabled?: boolean;
+		boldRanges?: Array<{ start: number; end: number }>;
+		highlightParts?: Array<{ text: string; isInputMatch: boolean }>;
+	}[] = [];
```
smorrisods: The same type appears again!

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -244,6 +342,22 @@ export class OntarioSearchBox {
 		this.updateCaptionState(this.caption);
 	}
 
+	private syncFormValue(value: string) {
```
smorrisods, line 341: Even private methods should have JSDocs.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +396,664 @@ export class OntarioSearchBox {
 		);
 	}
 
+	private async handleAutocompleteInput(query: string) {
+		this.autocompleteQueryUpdated.emit({ query });
+
+		if (!this.autocomplete) return;
+
+		if (query.length < (this.minChars ?? 1)) {
```
smorrisods: This looks like it is enforcing a default that matches a value defined above.  I'm wanting to suggest that we use a constant, but Stencil's parser will use the constant as the default in the `README.md` and not the number directly I don't think.  These two items should be tied to prevent drift.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +396,664 @@ export class OntarioSearchBox {
...
+			this.updateSlotSuggestionVisibility('');
+			this.suggestions = [];
+			this.emitSuggestionsUpdated();
+			this.closeSuggestions();
```
smorrisods: I think we should move `this.closeSuggestions()` above `this.suggestions = []` so there isn't a render race condition that has the suggestions vanish in the UI before the suggestions UI is closed.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
smorrisods: Or `enableAutocomplete`, whichever matches our conventions.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +396,664 @@ export class OntarioSearchBox {
...
+			return;
+		}
+
+		if (this.hasSuggestionSlotContent) {
+			this.updateSlotSuggestionVisibility(query);
+			this.activeSuggestionIndex = -1;
```
smorrisods: This shows up in a number of places, let's move it to a named helper method so there is one code path for resetting the `activeSuggestionIndex` as well as it adds a nice bit of pros to the code for readability.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +396,664 @@ export class OntarioSearchBox {
...
+			this.hoveredSuggestionIndex = -1;
```
smorrisods: This shows up in a number of places, let's move it to a named helper method so there is one code path for resetting the `hoveredSuggestionIndex` as well as it adds a nice bit of pros to the code for readability.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +396,664 @@ export class OntarioSearchBox {
...
+			} else {
+				this.closeSuggestions();
+			}
+
+			this.emitSuggestionsUpdated();
+			return;
+		}
+
+		if (!this.getSuggestions) {
+			this.suggestions = [];
```
smorrisods, line 429: Let's add reset suggestions method as we do this a few times.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +396,664 @@ export class OntarioSearchBox {
...
+			clearTimeout(this.debounceTimer);
+		}
+
+		this.debounceTimer = window.setTimeout(async () => {
+			const expectedQuery = this.value ?? '';
+
+			try {
+				const results = await this.getSuggestions?.(expectedQuery);
+
+				if (expectedQuery !== (this.value ?? '')) {
```
smorrisods: Let's make a `const` of this default value to check against rather than relying on the default and the check to stay in sync with refactors.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +396,664 @@ export class OntarioSearchBox {
...
+					this.openSuggestions();
+				} else {
+					this.closeSuggestions();
+				}
+			} catch {
+				this.suggestions = [];
+				this.emitSuggestionsUpdated();
+				this.closeSuggestions();
+			}
+		}, this.debounceMs ?? 150);
```
smorrisods: This shares the same default value as the property definition.  Similar to before, let's consider how to share a default value to avoid drift.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +396,664 @@ export class OntarioSearchBox {
...
+	}
+
+	private normalizeSuggestion(
```
smorrisods: JSDoc missing.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +396,664 @@ export class OntarioSearchBox {
...
+			| {
+					id?: string;
+					label: string;
+					value?: string;
+					description?: string;
+					href?: string;
+					disabled?: boolean;
+					boldRanges?: Array<{ start: number; end: number }>;
+					highlightParts?: Array<{ text: string; isInputMatch: boolean }>;
+			  },
```
smorrisods: The same type redefined anonymously again!

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +396,664 @@ export class OntarioSearchBox {
...
+		index: number,
+	) {
```
smorrisods: There is no return type defined for this method.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +396,664 @@ export class OntarioSearchBox {
...
+	private getAllSlotSuggestionElements(slot = this.suggestionSlotRef): HTMLElement[] {
+		const assignedElements = slot?.assignedElements({ flatten: true }) || [];
+		return assignedElements as HTMLElement[];
+	}
+
+	private getSlotSuggestionElements(slot = this.suggestionSlotRef): HTMLElement[] {
+		return this.getAllSlotSuggestionElements(slot).filter((el) => !el.hasAttribute('hidden'));
+	}
+
+	private updateSuggestionSlotState(slot?: HTMLSlotElement) {
```
smorrisods, line 531: No JSDocs on these methods, let's add them.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +396,664 @@ export class OntarioSearchBox {
...
+		const assignedOptions = this.getAllSlotSuggestionElements(slot);
+		this.hasSuggestionSlotContent = assignedOptions.length > 0;
+		this.updateSlotSuggestionVisibility(this.value || '', assignedOptions);
+		this.decorateSlotSuggestionOptions(this.getSlotSuggestionElements(slot));
+	}
+
+	private getSlotOptionLabel(option: HTMLElement): string {
+		if (option.tagName === 'ONTARIO-SEARCH-RESULT-ITEM') {
```
smorrisods: What's this magic string? 🪄 It should be a const that can be easily referenced wherever it is needed.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +396,664 @@ export class OntarioSearchBox {
...
+			}
+		});
+
+		if (this.activeSuggestionIndex >= this.getSuggestionCount()) {
+			this.activeSuggestionIndex = -1;
+		}
+
+		if (this.hoveredSuggestionIndex >= this.getSuggestionCount()) {
+			this.hoveredSuggestionIndex = -1;
+		}
```
smorrisods, line 568: The resets appear again.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +396,664 @@ export class OntarioSearchBox {
...
+	}
+
+	private computeFallbackHighlightParts(label: string, query: string): Array<{ text: string; isInputMatch: boolean }> {
```
smorrisods: This type (`{ text: string; isInputMatch: boolean }`) comes up a few times in this method, consider defining it as a `type`.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +396,664 @@ export class OntarioSearchBox {
...
+
+		if (queryIndex < queryChars.length) {
+			matchedIndices.clear();
+			const contiguousMatchStart = label.toLowerCase().indexOf(normalizedQuery.toLowerCase());
+			if (contiguousMatchStart >= 0) {
+				for (let i = contiguousMatchStart; i < contiguousMatchStart + queryChars.length; i++) {
+					matchedIndices.add(i);
+				}
+			}
+		}
```
smorrisods: I've paused my review here an I'll submit my responses so they don't linger in review.

smorrisods (changes requested):
This component set is genuinely interesting.  It's complexity is on the high side compared to many of our other components.  I've added some comments but I've paused my review a the respective comment as I try and understand how these components are working individually and together.  The addition of a slot also has increased the complexity with more nuances than I originally thought. Good work so far, @muntalee.

packages/ontario-design-system-component-library/src/index.html
muntalee: Fixed the indentation in ec5140a2

packages/ontario-design-system-component-library/src/components/ontario-search-box/readme.md
muntalee: Added React and Angular tabs to the async suggestions example in ba3b8a4b. Both frameworks show how to wire up `getSuggestions` as a callback prop/binding

packages/ontario-design-system-component-library/src/components/ontario-search-box/readme.md
muntalee: I split "Important notes about autocomplete" into three separate paragraphs in ba3b8a4b so each note renders distinctly in the docs.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Renamed `enableAutocomplete` in 570e3a17. Matches the enable prefix convention used elsewhere in the codebase. Updated throughout the component, tests, index.html, and readme.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Extracted as `export interface AutocompleteSuggestion` in e73eafa8. It's exported so it's available in `components.d.ts` for consumers of the library.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: It's the accessible label for the search input, so it must always be provided. The `!` non-null assertion is intentional and consistent with how other required props like `inputOnInput` are declared throughout the codebase.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: The `!` is a TypeScript non-null assertion used by Stencil to declare that Stencil will always initialise the `EventEmitter` before use, not that callers are required to handle it. This is the standard pattern for `@Event()` declarations throughout the component library (see `ontario-input`, `ontario-checkboxes`, etc.).

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Ooo good catch. The `autocompleteSuggestionSelected` event payload now uses `AutocompleteSuggestionSelectedEvent` (a named exported interface) in 2b673bb0. This also fixed a CI build failure where Stencil's Angular wrapper generator was serialising the inline object type as `[object Object]`.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Resolved in e73eafa8, so the `@State()` private suggestions array is now typed as `AutocompleteSuggestion[]` using the named interface.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Added JSDoc to all private methods in dcd4592d, including `syncFormValue`, `normalizeSuggestion`, `getSuggestionListId`, `getAriaLiveRegionId`, `getAllSlotSuggestionElements`, `getSlotSuggestionElements`, `getSuggestionCount`, and `isSuggestionDisabled`.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Extracted `DEFAULT_MIN_CHARS = 1`, `DEFAULT_DEBOUNCE_MS = 150`, and `DEFAULT_MAX_SUGGESTIONS = 8` as private static constants in 763263d5. All inline literals that checked against those defaults now reference the constants.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Fixed in 763263d5, where `closeSuggestions()` is now called before `suggestions = []` in all three code paths (`below-min-chars`, `no-getSuggestions`, and catch block)

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Added `resetActiveSuggestionIndex()` in dcd4592d. All direct assignments to -1 have been replaced with the helper call.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Added `resetHoveredSuggestionIndex()` in dcd4592d, same approach as above.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Added `resetSuggestions()` in dcd4592d, which closes the suggestion list and clears the array in the correct order.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Updated the JSDoc for `debounceMs` to read "Debounce delay in milliseconds before getSuggestions` is called" in dcd4592d.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Replaced `?? 150` with `?? OntarioSearchBox.DEFAULT_DEBOUNCE_MS` in 763263d5.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Added JSDoc to `normalizeSuggestion` in dcd4592d.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Resolved in e73eafa8 — the parameter type is now `string | AutocompleteSuggestion`, and the return type is `AutocompleteSuggestion`.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Added `AutocompleteSuggestion` return type in e73eafa8.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Covered in the same dcd4592d commit as above.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Extracted as `private static readonly RESULT_ITEM_TAG = 'ONTARIO-SEARCH-RESULT-ITEM'` in 763263d5. All usages now reference the constant.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Resolved by the `resetActiveSuggestionIndex()` and `resetHoveredSuggestionIndex()` helpers from dcd4592d — all remaining inline assignments have been replaced.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Extracted as `type HighlightPart = { text: string; isInputMatch: boolean }` at module level in 763263d5. `computeFallbackHighlightParts` now returns `HighlightPart[]` and uses it for internal variables.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Oops missed this one! Fixed in bf06466b — the staleness check now uses `const currentQuery = this.value ?? ''` so both sides of the comparison share the same normalisation pattern. If the default for `value` ever changes, there's only one place to update.

packages/ontario-design-system-component-library/src/components/ontario-search-box/test/ontario-search-box.e2e.ts
```diff
@@ -1,21 +1,65 @@
...
+
+		await page.evaluate(() => {
+			const searchBox = document.querySelector('ontario-search-box') as HTMLOntarioSearchBoxElement;
+			searchBox.getSuggestions = async () => ['Toronto', 'Ottawa', 'London'];
+			searchBox.debounceMs = 0;
+		});
+
+		const input = page.locator('ontario-search-box').locator('input[type="search"]');
+		await input.fill('to');
+		await page.waitForTimeout(20);
```
erincarrieresontario: I think for this, instead of waiting for an arbitrary amount of time, we should wait for a condition

packages/ontario-design-system-component-library/src/components/ontario-search-box/test/ontario-search-box.e2e.ts
```diff
@@ -1,21 +1,65 @@
...
+	});
+
+	test('uses slot suggestions when provided', async ({ page }) => {
+		await page.setContent(`<ontario-search-box enable-autocomplete caption="Search cities">
+			<ontario-search-result-item slot="suggestions" label="Ajax" value="Ajax"></ontario-search-result-item>
+			<ontario-search-result-item slot="suggestions" label="Barrie" value="Barrie"></ontario-search-result-item>
+		</ontario-search-box>`);
+
+		const input = page.locator('ontario-search-box').locator('input[type="search"]');
+		await input.fill('a');
```
erincarrieresontario, line 37: Hmm for this test, I don't know if we are actually testing if the autocomplete suggestions actually exist, because the component could completely fail to render autocomplete suggestions and this test would still pass. I think instead we should check for the autocomplete UI

packages/ontario-design-system-component-library/src/components/ontario-search-box/test/ontario-search-box.spec.tsx
```diff
@@ -1,15 +1,170 @@
 import { render } from '@stencil/vitest';
```
erincarrieresontario, line 1: These tests look good! I think we could also cover cases for ArrowDown, ArrowUp, Enter and Escape

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.scss
```diff
@@ -136,6 +142,125 @@ input[type='reset'].ontario-search__reset {
...
+	background-color: colours.$ontario-colour-white;
+	border-radius: 0 0 globalVariables.$global-radius globalVariables.$global-radius;
+	border-top: globalFunctions.px-to-rem(2) solid $ontario-search-autosuggest-box-shadow;
+	box-shadow: 0 globalFunctions.px-to-rem(3) globalFunctions.px-to-rem(8) globalFunctions.px-to-rem(1)
+		$ontario-search-autosuggest-box-shadow;
+	clip-path: inset(0 globalFunctions.px-to-rem(-8) globalFunctions.px-to-rem(-8) globalFunctions.px-to-rem(-8));
+	display: none;
+	list-style: none;
+	margin: globalFunctions.px-to-rem(-2) 0 0 0;
+	padding: 0;
```
erincarrieresontario, line 155: For where we use zero, we can use a variable for this: `spacing.$spacing-0`

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.scss
```diff
@@ -136,6 +142,125 @@ input[type='reset'].ontario-search__reset {
...
+		outline: none;
+	}
+}
+
+.ontario-search-autocomplete__suggestion-option:last-child {
+	padding-bottom: spacing.$spacing-2;
+}
+
+.ontario-search-autocomplete__suggestion-option--active {
+	background-color: colours.$ontario-colour-button-tertiary--hover;
```
erincarrieresontario, line 188: This is interesting, are we adding action classes using JS? we should be able to just target hover and active within CSS without adding the classes. Or are we adding them for another reason?

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.scss
```diff
@@ -136,6 +142,125 @@ input[type='reset'].ontario-search__reset {
...
+::slotted(.ontario-search-autocomplete__slot-option--hovered),
+::slotted([slot='suggestions']:hover) {
+	background-color: colours.$ontario-colour-button-tertiary--hover;
+}
+
+::slotted(.ontario-search-autocomplete__slot-option) {
+	display: block;
+	list-style: none;
+	padding: spacing.$spacing-2 spacing.$spacing-4;
+	margin: 0;
```
erincarrieresontario: Can use a variable here

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.scss
```diff
@@ -136,6 +142,125 @@ input[type='reset'].ontario-search__reset {
...
+	line-height: lineheight.$ontario-line-height-8;
+	background-color: colours.$ontario-colour-white;
+	cursor: pointer;
+	box-sizing: border-box;
+	font-weight: fontWeights.$ontario-font-weights-bold;
+}
+
+::slotted([slot='suggestions'][hidden]),
+::slotted(.ontario-search-autocomplete__slot-option[hidden]) {
+	display: none !important;
```
erincarrieresontario, line 237: Lets add a comment as to why we are using the important tag here

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.scss
```diff
@@ -136,6 +142,125 @@ input[type='reset'].ontario-search__reset {
...
+
+.ontario-search-autocomplete__suggestion-description {
+	line-height: 1.35;
+}
+
+.ontario-search__visually-hidden {
+	position: absolute;
+	width: globalFunctions.px-to-rem(1);
+	height: globalFunctions.px-to-rem(1);
+	margin: -1px;
```
erincarrieresontario: This seems like a magic number, lets add a comment and also a px-to-rem

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -327,39 +1137,105 @@ export class OntarioSearchBox {
...
+										.filter(Boolean)
+										.join(' ')}
+									data-ontario-suggestion-index={String(index)}
+									data-value={suggestion.value || suggestion.label}
+									role="option"
+									aria-selected={String(index === this.activeSuggestionIndex || index === this.hoveredSuggestionIndex)}
+									aria-disabled={String(!!suggestion.disabled)}
+									aria-label={suggestion.label}
+									tabIndex={-1}
+									onClick={(event) => {
```
erincarrieresontario: Both onMouseDown and onClick are calling `selectSuggestionByIndex`, wondering if we should just call it on mouseDown in order to not call it twince 🤔

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +387,697 @@ export class OntarioSearchBox {
...
+		);
+	}
+
+	private handleSuggestionMouseOver = (event: MouseEvent) => {
+		if (!this.enableAutocomplete) return;
+
+		const hoveredIndex = this.getSuggestionIndexFromEvent(event);
+		if (hoveredIndex < 0 || hoveredIndex === this.hoveredSuggestionIndex) return;
+
+		this.hoveredSuggestionIndex = hoveredIndex;
```
erincarrieresontario, line 742: Are we clearing this hover state anywhere?

erincarrieresontario (changes requested):
Overall this looks really strong @muntalee , just some questions and comments for you to take a look at :)

packages/ontario-design-system-component-library/src/components/ontario-search-box/test/ontario-search-box.e2e.ts
muntalee: I replaced the fixed wait with condition-based assertions in 42a80aaa to make the test wait for the autocomplete UI instead of an arbitrary delay.

packages/ontario-design-system-component-library/src/components/ontario-search-box/test/ontario-search-box.e2e.ts
muntalee: Ooo good catch. I updated the e2e coverage in 42a80aaa to assert on the rendered autocomplete UI/result items, so the test fails if suggestions do not appear.

packages/ontario-design-system-component-library/src/components/ontario-search-box/test/ontario-search-box.spec.tsx
muntalee: I expanded the unit coverage in 42a80aaa to cover ArrowDown, ArrowUp, Enter, and Escape behaviour. I'll have to investigate the other test branch for this component and see if any other details are needed.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.scss
muntalee: I replaced the raw zero values with `spacing.$spacing-0` in 68cf0535 to keep the scss consistent with the design tokens.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.scss
muntalee: Added in 68cf0535!

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.scss
muntalee: Valid. In 68cf0535 I added one stating we need the !important there so consumer slot styles cannot override the hidden state.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.scss
muntalee: I changed that offset to `px-to-rem(-1)` and added a comment in 68cf0535 explaining why it is there.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: I removed the duplicate click path in 67b83ba1 and kept selection on mousedown so the suggestion is only selected once.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Yup! The hover state is reset when suggestions close in 67b83ba1, so we do not leave stale hover styling behind.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.scss
muntalee: Since the autocomplete refactor happened after this comment, the rendering path changed quite a bit. In 67b83ba1 I kept the active/hover classes so async suggestions would keep the same visual behaviour after that rework. I'm more than welcome to revisit and try a more CSS-only approach.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +388,601 @@ export class OntarioSearchBox {
...
+			option.classList.add('ontario-search-autocomplete__slot-option');
+
+			const isActive = index === this.activeSuggestionIndex || index === this.hoveredSuggestionIndex;
+			const isHovered = index === this.hoveredSuggestionIndex;
+			option.setAttribute('aria-selected', String(isActive));
+			// Slot options need explicit state classes so keyboard navigation can style active rows without relying on :hover.
+			option.classList.toggle('ontario-search-autocomplete__slot-option--active', isActive);
+			option.classList.toggle('ontario-search-autocomplete__slot-option--hovered', isHovered);
+
+			if (option.tagName === 'ONTARIO-SEARCH-RESULT-ITEM') {
```
smorrisods: This checks `option.tagName === 'ONTARIO-SEARCH-RESULT-ITEM'` as a literal string, but you just defined `OntarioSearchBox.RESULT_ITEM_TAG` for exactly this a few lines up (used correctly in `getSlotOptionLabel`). Let's use the constant here too so the two checks can't drift out of sync.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +388,601 @@ export class OntarioSearchBox {
...
+			description: item.description,
+			href: item.href,
+			disabled: !!item.disabled,
+			segments: this.resolveSuggestionSegments(label, query, item),
+			boldRanges: item.boldRanges,
+			highlightParts: item.highlightParts,
+		};
+	}
+
+	private resolveSuggestionSegments(label: string, query: string, suggestion?: AutocompleteSuggestion): Segment[] {
```
smorrisods: `resolveSuggestionSegments` here and `getResolvedSegments` in `ontario-search-result-item.tsx` are doing the same `segments` → `highlightParts` → `boldRanges` → fallback resolution, nearly line-for-line. Since `search-box-autocomplete.ts` already exists as the shared util for this feature, let's move this resolution logic there once and have both components call it, so there's one code path for segment fallback behaviour.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +388,601 @@ export class OntarioSearchBox {
 		);
 	}
 
+	private async handleAutocompleteInput(query: string) {
```
smorrisods, line 398: None of the new private methods have JSDocs (`handleAutocompleteInput`, `resolveSuggestionSegments`, `updateSuggestionSlotState`, `selectSuggestionByIndex`, `handleInputKeyDown`, etc. — this applies broadly across the file). Even though they're private, they should still have doc comments; a handful of the older/existing helpers already do (`getSuggestionListId`, `resetSuggestions`), so let's bring the new ones up to the same standard.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +388,601 @@ export class OntarioSearchBox {
...
+		this.closeSuggestions();
+		this.suggestions = [];
+	}
+
+	private selectSuggestionByIndex(index: number, source: 'keyboard' | 'mouse'): void {
+		if (index < 0 || this.isSuggestionDisabled(index)) {
+			return;
+		}
+
+		let selectedSuggestion:
```
smorrisods: This redeclares an anonymous type that's effectively `AutocompleteSuggestion` (already defined at the top of the file). Let's use `AutocompleteSuggestion` directly instead of a fresh inline shape.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -255,8 +361,14 @@ export class OntarioSearchBox {
...
-		if (typeof this.internals?.setFormValue === 'function') {
-			this.internals.setFormValue(this.value ?? '');
+		this.syncFormValue(this.value ?? '');
+
+		if (eventType === EventType.Input) {
+			this.handleAutocompleteInput(this.value ?? '');
+		}
+
+		if (eventType === EventType.Blur && this.enableAutocomplete) {
+			window.setTimeout(() => this.closeSuggestions(), 120);
```
smorrisods: `window.setTimeout(() => this.closeSuggestions(), 120)` — this magic `120` isn't a `const` like the other new defaults (`DEFAULT_MIN_CHARS`, `DEBOUNCE_MS`, etc.). Worth naming it (e.g. `BLUR_CLOSE_DELAY_MS`) and a short comment on why it exists — presumably to let a suggestion's `mousedown` fire before blur closes the list?

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +388,601 @@ export class OntarioSearchBox {
...
+
+	private getSlotOptionLabel(option: HTMLElement): string {
+		if (option.tagName === OntarioSearchBox.RESULT_ITEM_TAG) {
+			return (option as any).label || (option as any).value || this.getSuggestionValueFromOption(option);
+		}
+
+		return this.getSuggestionValueFromOption(option);
+	}
+
+	private updateSlotSuggestionVisibility(query: string, assignedOptions = this.getAllSlotSuggestionElements()) {
```
smorrisods: Is subsequence/fuzzy matching intentional here for filtering slotted options (via `computeHighlightSegments`), or should slot filtering just require the query as a contiguous substring? Worth confirming since it affects which slotted suggestions get hidden as the user types.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -15,9 +15,32 @@ import {
...
+/** Represents a suggestion item in autocomplete mode */
+export interface AutocompleteSuggestion {
+	id?: string;
+	label: string;
+	value?: string;
+	description?: string;
+	href?: string;
+	disabled?: boolean;
+	segments?: Segment[];
+	boldRanges?: Array<{ start: number; end: number }>;
```
smorrisods, line 32: `highlightParts`/`boldRanges` are marked deprecated on `ontario-search-result-item.tsx`, but these matching fields on `AutocompleteSuggestion` aren't. Let's mirror the `@deprecated` note here too for consistency.

smorrisods (changes requested):
Great follow-up — the `enableAutocomplete` naming and the new `DEFAULT_*` constants address the earlier feedback nicely. A few things before this merges: some duplicated segment-resolution logic between the two components, missing JSDocs on the new private helpers, and one unrelated readme diff that's probably not meant to be here. Also flagged one small drift risk (hardcoded tag string) and one open question on slot-filtering behaviour.

---
🤖 A robot wrote this in Scott's voice. It does not, in fact, have opinions about `const`s — Scott does.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +388,601 @@ export class OntarioSearchBox {
...
+		return this.getSuggestionValueFromOption(option);
+	}
+
+	private updateSlotSuggestionVisibility(query: string, assignedOptions = this.getAllSlotSuggestionElements()) {
+		const normalizedQuery = (query || '').trim();
+		const hasQuery = normalizedQuery.length > 0;
+
+		assignedOptions.forEach((option) => {
+			const optionLabel = this.getSlotOptionLabel(option);
+			const shouldShow = !hasQuery || computeHighlightSegments(optionLabel, normalizedQuery) !== null;
```
smorrisods: Per the design doc for this feature, the slot case (Case 4) is supposed to be fully consumer-owned — we don't reach in and decide relevance for projected content, regardless of markup. This line has us auto-hiding slotted options based on `computeHighlightSegments(...) !== null`, which means the component *is* making a relevance decision on the consumer's slotted content. That contradicts the "no reach-in" rule for the slot path. Should this filtering be removed here and left to the consumer to implement themselves (using the exported `computeHighlightSegments`) if they want it, same as Case 4's `addEventListener('autocompleteQueryUpdated', ...)` example shows?

smorrisods (changes requested):
Checked this against the API design doc this implementation is meant to follow — matching-order fix (contiguous-first, fuzzy-fallback) lines up correctly, and the async `getSuggestions` path doesn't drop candidates on a `null` match either, which is good. One thing doesn't line up though, flagged inline.

---

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Fixed in 330cf69 by using the shared result-item tag constant in both places.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Fixed in 5f299983 by moving the shared fallback logic into search-box-autocomplete.ts.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Keep on missing these 😭. Fixed in 330cf69 by adding docs to the new private helpers.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Fixed in 330cf69 by using AutocompleteSuggestion directly.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Fixed in 330cf69 by naming the delay constant and documenting why it exists.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: I updated this in 330cf69 so the slot case stays consumer-owned and no longer filters projected content itself.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Fixed in 330cf69 by mirroring the deprecation notes onto AutocompleteSuggestion.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Good call! On 330cf69 we removed component-side slot relevance filtering (`computeHighlightSegments(...) !== null`) so the slot path no longer decides which projected options are relevant; consumers now own that logic and we only mirror visibility/accessibility state.

PR #303, opened 2026-07-14
DS-2703: Implement `ontario-search-result-item` component

muntalee:
## Summary
Creates the new `ontario-search-result-item` Stencil component to provide a reusable, semantic option for autocomplete and search suggestion results in the `ontario-search-box` component.

## Merge Notice
This PR should only be merged if [DS-2682](https://github.com/ongov/ontario-design-system/pull/322) and [DS-2683](https://github.com/ongov/ontario-design-system/pull/304) are merged in this PR.

## Changes
- New component: `ontario-search-result-item` with unit tests and documentation
- Support for optional label/description/href/value props
- Support for slotted custom content
- Proper ARIA semantics for listbox option role
- React wrapper exports
- Angular wrapper exports
- Modified `index.html` for developer preview


packages/ontario-design-system-component-library/src/components/ontario-search-result-item/test/ontario-search-result-item.spec.tsx
erincarrieresontario, line 1: This looks good, just a few more test cases to check for: 

- it('should not emit itemSelected when disabled')
- it('should render href when provided')
- it('should update aria-selected when selected changes')
Show more lines

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.scss
```diff
@@ -0,0 +1,65 @@
...
+	cursor: default;
+	opacity: 0.65;
+}
+
+:host(:focus-visible) .ontario-search-result-item {
+	box-shadow: 0 0 0 globalFunctions.px-to-rem(4) colours.$ontario-colour-focus;
+	outline: globalFunctions.px-to-rem(4) solid transparent;
+}
+
+.ontario-search-result-item__content {
```
erincarrieresontario: Is this style needed?

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
```diff
@@ -0,0 +1,221 @@
...
+	@Prop() language?: Language = 'en';
+
+	/**
+	 * Emitted when a non-disabled option is selected via click.
+	 */
+	@Event() itemSelected!: EventEmitter<{ label?: string; value?: string; href?: string }>;
+
+	@State() private hasDefaultSlot = false;
+
+	@Watch('active')
```
erincarrieresontario: One area worth revisiting is the use of @Watch handlers to imperatively update host attributes such as aria-selected and aria-disabled. Since these values are derived directly from component state/props, it may be cleaner and more maintainable to bind them on the <Host> element during render rather than mutating the DOM via setAttribute(). This keeps the rendered output as the single source of truth, reduces lifecycle complexity, and aligns more closely with typical Stencil patterns.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
```diff
@@ -0,0 +1,221 @@
...
+			</span>
+		);
+	}
+
+	render() {
+		const value = this.value || this.label;
+		const labelText = this.label || '';
+
+		return (
+			<Host class={this.getHostClassNames()} onClick={this.onSelect} data-value={value}>
```
erincarrieresontario: One potential concern is the interaction between the clickable host element and the optional href link. Clicking a linked result may trigger both the component's itemSelected event and the browser's navigation behaviour at the same time. It would be worth confirming that this is the intended behaviour and that consumers have an opportunity to handle the selection event before navigation occurs.

erincarrieresontario (changes requested):
Looks good @muntalee , just a few comments :)

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/test/ontario-search-result-item.spec.tsx
```diff
@@ -0,0 +1,76 @@
...
+	it('should render description when provided', async () => {
+		const page = await render(
+			`<ontario-search-result-item label="Toronto" description="City in Ontario"></ontario-search-result-item>`,
+		);
+		await page.waitForChanges();
+
+		const description = (page.root as HTMLElement).shadowRoot?.querySelector(
+			'.ontario-search-result-item__description',
+		);
+		expect(description?.textContent).toContain('City in Ontario');
```
smorrisods: The string `City in Ontario` can be made a `const` and couple the input and the output to verify the test is working and maintainable.

There are other instances here where this can be done.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.scss
```diff
@@ -0,0 +1,65 @@
...
+.ontario-search-result-item__match {
+	font-weight: fontWeights.$ontario-font-weights-normal;
+}
+
+.ontario-search-result-item__completion {
+	font-weight: fontWeights.$ontario-font-weights-bold;
+}
+
+.ontario-search-result-item__description {
+	line-height: 1.35;
```
smorrisods: Let's name this magic number, or if it matches a default value let's use that instead.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
```diff
@@ -0,0 +1,221 @@
...
+	 * Optional highlight parts used to style input-matched text and completion text.
+	 * When provided, this takes precedence over default label rendering.
+	 */
+	@Prop() highlightParts?: Array<{ text: string; isInputMatch: boolean }>;
+
+	/**
+	 * Optional bold ranges over the label string for completion emphasis.
+	 * Used when highlightParts is not provided.
+	 */
+	@Prop() boldRanges?: Array<{ start: number; end: number }>;
```
smorrisods: As per the architectural redesign these aren't needed anymore and should be restructured into `segments`.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	componentDidLoad() {
+		const slotElement = this.element.shadowRoot?.querySelector('slot');
+		if (slotElement) {
+			this.syncSlotState(slotElement as HTMLSlotElement);
+		}
+	}
+
+	private syncSlotState(slotElement: HTMLSlotElement) {
```
smorrisods, line 83: These methods are missing JSDocs, we should add them.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
```diff
@@ -0,0 +1,221 @@
...
+			'ontario-search-result-item',
+			this.active && 'ontario-search-result-item--active',
+			this.selected && 'ontario-search-result-item--selected',
+			this.disabled && 'ontario-search-result-item--disabled',
+		]
+			.filter(Boolean)
+			.join(' ');
+	}
+
+	private getResolvedHighlightParts(label: string): Array<{ text: string; isInputMatch: boolean }> {
```
smorrisods: This method will probably be replaced when the changes for segments are implemented.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/readme.md
```diff
@@ -0,0 +1,66 @@
...
+```html
+<ontario-search-result-item value="Toronto">
+	<div>
+		<strong>Toronto</strong>
+		<p>Population: 2.9M</p>
+	</div>
+</ontario-search-result-item>
+```
+
+<!-- Auto Generated Below -->
```
smorrisods, line 70: There's nothing autogenerated here! The build should be run so the readme updates.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/readme.md
```diff
@@ -0,0 +1,66 @@
+import { OntarioSearchResultItem } from '@ongov/ontario-design-system-component-library-react';
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# ontario-search-result-item
+
+Use `ontario-search-result-item` to render a semantic option row for search suggestions.
```
smorrisods, line 7: The beginning of this readme is missing the usual fanfare of referencing the guidance for the component it is part of.

smorrisods (changes requested):
Good work, @muntalee.  I've added some comment around consistency and refactoring.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/test/ontario-search-result-item.spec.tsx
muntalee: Added the missing test coverage in c88de32f so that disabled selection, href rendering, and aria-selected updates are all explicitly validated.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.scss
muntalee: Kept the focus outline rule and documented why in ca64b8b4. The outline is there to preserve a visible keyboard focus state, especially in high-contrast scenarios where the shadow alone is not enough.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
muntalee: Refactored that logic in 858d367e so aria-selected and aria-disabled are now derived directly in render instead of being set imperatively through watchers. That makes the host state easier to reason about and keeps the markup consistent with component props.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
muntalee: Adjusted the click handling in 858d367e so the link and host no longer compete. The link now stops propagation before selection logic runs, which keeps navigation and itemSelected emission from firing twice.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/test/ontario-search-result-item.spec.tsx
muntalee: Cleaned up the test data in c88de32f by pulling the repeated strings into shared constants. That keeps the assertion tied to the input value and makes the tests easier to maintain when the sample text changes.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.scss
muntalee: Named the description line-height in 858d367e by moving the value into a SCSS variable. That removes the hard-coded number and makes the styling intent clearer.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
muntalee: Updated the component API in 858d367e to use segments instead of highlightParts and boldRanges. The render path now consumes the new segment shape directly, which matches the autocomplete branch design.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
muntalee: Added JSDoc to the slot and rendering helpers in 858d367e

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
muntalee: Replaced the old highlight resolver in 858d367e with the segments-based implementation, so the component now follows the new highlighting model end to end.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/readme.md
muntalee: Woops, Regenerated the component docs in ca64b8b4

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/readme.md
muntalee: Fixed the README introduction in 8a8dbc07 by adding the component guidance reference.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
```diff
@@ -0,0 +1,217 @@
...
+		if (slotElement) {
+			this.syncSlotState(slotElement as HTMLSlotElement);
+		}
+	}
+
+	/**
+	 * Synchronise whether fallback text should render based on assigned slot nodes.
+	 */
+	private syncSlotState(slotElement: HTMLSlotElement) {
+		this.hasDefaultSlot = !!slotElement.assignedNodes({ flatten: true }).length;
```
erincarrieresontario: Consider using assignedElements() instead of assignedNodes() if the goal is to detect actual slotted content. Because assignedNodes() may treat whitespace text nodes as content I belieev

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
```diff
@@ -0,0 +1,217 @@
...
+		const labelText = this.label || '';
+		const isSelected = !!(this.active || this.selected);
+		const isDisabled = !!this.disabled;
+
+		return (
+			<Host
+				role="option"
+				tabIndex={-1}
+				aria-selected={String(isSelected)}
+				aria-disabled={String(isDisabled)}
```
erincarrieresontario: It might be more readable to do something like this: 

aria-selected={isSelected ? 'true' : 'false'}
aria-disabled={isDisabled ? 'true' : 'false'}

erincarrieresontario (approved):
Looks great @muntalee , just a few comments :)

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
muntalee: Good catch! Switched to `assignedElements()` so whitespace-only text nodes between tags don't falsely trigger `hasDefaultSlot` in dc7fc2e.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
muntalee: Love that, changed it in dc7fc2e.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.scss
```diff
@@ -0,0 +1,64 @@
...
+	cursor: pointer;
+}
+
+.ontario-search-result-item--active,
+.ontario-search-result-item--selected {
+	background-color: colours.$ontario-colour-button-tertiary--hover;
+}
+
+.ontario-search-result-item--selected {
+	box-shadow: 0 0 0 globalFunctions.px-to-rem(4) colours.$ontario-colour-focus inset;
```
smorrisods: `globalFunctions.px-to-rem(4)` shows up three times here for the focus ring width. If that value ever needs to change, it's easy to update two of the three and miss one. Let's pull it into a `$search-result-item-focus-ring-width` (or similar) `scss` variable once and reference it in all three spots.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+	/**
+	 * Emit a selection event when interaction is allowed.
+	 */
+	private emitSelection() {
+		if (this.disabled) return;
+
+		this.itemSelected.emit({
+			label: this.label,
+			value: this.value || this.label,
```
smorrisods: We compute `this.value || this.label` in both `emitSelection()` and `render()`. Let's pull this into a small private getter (e.g. `private get resolvedValue()`) so the fallback logic only lives in one place — if the precedence ever changes we won't have to remember both spots.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
```diff
@@ -0,0 +1,217 @@
...
+	 * Optional language prop to align with component API conventions.
+	 */
+	@Prop() language?: Language = 'en';
+
+	/**
+	 * Emitted when a non-disabled option is selected via click.
+	 */
+	@Event() itemSelected!: EventEmitter<{ label?: string; value?: string; href?: string }>;
+
+	@State() private hasDefaultSlot = false;
```
smorrisods, line 71: No JSDoc on `@State() private hasDefaultSlot`. Let's add one describing what it tracks (whether the default slot has projected content) — even private/internal state benefits from a one-liner for the next person reading this.

smorrisods (changes requested):
Nice work on this one — the component is clean and well tested. A few small notes: a repeated magic value in the scss, a small duplicated fallback in the tsx, and one missing JSDoc on internal state.

---
🤖 Beep boop, this is what Scott would say if Scott were 40% more caffeinated and 100% synthetic.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.scss
muntalee: Gotchu, pulled the `4px` value into `$search-result-item-focus-ring-width` and replaced all three occurrences in dd777c6.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
muntalee: Added a `private get resolvedValue()` getter and replaced both `this.value || this.label` usages with it in dd777c6.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
muntalee: Added a one-liner: `/** Whether the default slot has any projected element content. */`
\
dd777c6

smorrisods (approved):
LGTM 🚀

PR #304, opened 2026-07-14
DS-2683: Add comprehensive unit and end-to-end tests for search-box autocomplete

muntalee:
## Summary
Adds comprehensive unit and end-to-end test coverage for the autocomplete functionality in `ontario-search-box`, including async suggestions, slotted content, keyboard navigation, accessibility validation, and regression safety.

## Merge Notice
This PR is dependent on #322

## Changes
- Unit tests
  - Autocomplete disabled by default
  - Async suggestions: fetching, rendering, error handling
  - Slotted content: rendering, slot change handling, precedence over async
  - Keyboard navigation: arrow keys, Enter, Escape
  - Mouse selection and interaction
  - Event emissions (query-changed, suggestions-updated, suggestion-selected)
  - Input value updates on selection
  - Backward compatibility with existing search functionality

- E2E Tests
  - Full typing and suggestion flow workflows
  - Keyboard and mouse selection workflows
  - Accessibility validation (roles, attributes, focus behaviour)
  - Screen reader announcement testing
  - Cross-browser compatibility checks
- VRT
  - Default, required, hint text, pre-filled value, French language, and caption variant states
  - Input, submit button, and reset button focus states
  - Autocomplete idle state
  - Slotted suggestions list open, first and second item keyboard highlighted
  - Async suggestions list open and first item keyboard highlighted
  - Partial match highlight segment rendering

- Component fetches suggestions via `getSuggestions(query)` callback
- Developers slot `ontario-search-result-item` children; component manages interaction
- Developers slot arbitrary HTML; component manages open/close/selection


packages/ontario-design-system-component-library/src/components/ontario-search-box/test/ontario-search-box.spec.tsx
```diff
@@ -60,6 +60,98 @@ describe('ontario-search-box', () => {
...
+		) as HTMLInputElement;
+		input.value = 'wat';
+		input.dispatchEvent(new Event('input', { bubbles: true, composed: true }));
+		await page.waitForChanges();
+
+		const semanticOptions = (page.root as HTMLElement).querySelectorAll(
+			'ontario-search-result-item[slot="suggestions"]',
+		);
+		expect((semanticOptions[0] as HTMLElement).hasAttribute('hidden')).toBe(true);
+		expect((semanticOptions[1] as HTMLElement).hasAttribute('hidden')).toBe(false);
```
smorrisods: Is this pattern deterministic?  We're checking that specific elements in a specific order are intended to have certain properties without validating the element being asserted against is the element we intend to have the intended property.

smorrisods (changes requested):
Loving the tests, this is looking good.  Only one minor nitpick.

packages/ontario-design-system-component-library/src/components/ontario-search-box/test/ontario-search-box.spec.tsx
muntalee: Valid, relying on index position wasn't deterministic. I've updated both tests to use attribute selectors instead in 21ce237.

packages/ontario-design-system-component-library/src/components/ontario-search-box/test/ontario-search-box.spec.tsx
```diff
@@ -59,6 +59,134 @@ describe('ontario-search-box', () => {
...
+			'ontario-search-result-item[slot="suggestions"][value="Toronto"]',
+		) as HTMLElement;
+
+		expect(waterlooOption.hasAttribute('hidden')).toBe(false);
+		expect(torontoOption.hasAttribute('hidden')).toBe(true);
+	});
+
+	it('should close suggestion list on mouse option selection', async () => {
+		const page = await render(`<ontario-search-box enable-autocomplete caption="Search cities"></ontario-search-box>`);
+		const hostRef = (page.root as any).__stencil__getHostRef?.();
```
erincarrieresontario: The use of __stencil__getHostRef() works, but it does couple the test to Stencil internals. If there's a straightforward way to trigger selection through a user interaction, that would be a more maintainable test long-term.

erincarrieresontario (commented):
Just one comment :)

packages/ontario-design-system-component-library/src/components/ontario-search-box/test/ontario-search-box.spec.tsx
muntalee: Removed `__stencil__getHostRef__()` and replaced it with dispatching a `mousedown` event directly on the rendered `ontario-search-result-item` in the shadow DOM in 0aa4e1c.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -536,11 +540,33 @@ export class OntarioSearchBox {
...
 	}
 
+	private doesSlotOptionMatchQuery(option: HTMLElement, query: string): boolean {
+		const trimmedQuery = (query || '').trim();
+		if (!trimmedQuery) {
+			return true;
+		}
+
+		const label = this.getSlotOptionLabel(option);
+		if (!label) {
```
smorrisods: Per the highlight-API design doc I shared, Case 4 (the `suggestions` slot) is explicitly "consumer owns composition" — the component doesn't decide relevance for slotted content, and if the consumer wants highlighting/filtering help they call `computeHighlightSegments` themselves.

`doesSlotOptionMatchQuery`/`updateSlotSuggestionVisibility` do the opposite of that: the component now calls `computeHighlightSegments` itself and sets/removes `hidden` on slotted content unconditionally. For our own `<ontario-search-result-item>` children this is low-risk since `label`/`value` are well-defined. But the slot is documented to accept *any* markup, and for that case the label is guessed via `getSlotOptionLabel`/`getSuggestionValueFromOption` (`data-value` attr → `value` attr/prop → raw `textContent`). That's a fragile heuristic once the slotted content is a rich composite (icon + label + description + price, etc.) — `textContent` concatenates all of it, so a match/non-match can easily be wrong for reasons that have nothing to do with the intended label. Worse, it's forced with no opt-out: a consumer who pre-filtered their own content server-side (Case 3-style — typo-tolerant/synonym results that don't literally contain the query substring) would now have the component silently re-hide entries they deliberately chose to show.

Since the design already settled this — the slot is consumer-owned, and the component should mirror whatever `hidden` state the consumer sets rather than deciding relevance itself — let's revert `updateSlotSuggestionVisibility` back to that (the `_query` param can stay unused, as it was before this PR) and drop `doesSlotOptionMatchQuery`. If we want to offer filtering-on-behalf-of-the-consumer as a feature, that should be its own explicit opt-in prop with its own design pass, not the default behaviour for every consumer using the slot today.

smorrisods (changes requested):
Great breadth of coverage here — the VRT and e2e suites are thorough. One thing needs to change before merge though: this isn't just tests, `updateSlotSuggestionVisibility` adds the component deciding relevance and hiding slotted content on the consumer's behalf, which contradicts the highlight-API design's Case 4 (the slot stays consumer-owned). Let's revert that part and keep the new coverage for everything else.

---

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Reworked in 4ff83ce to have slots stay consumer-owned

erincarrieresontario (approved):
LGTM 👍

packages/ontario-design-system-component-library/src/components/ontario-search-box/test/ontario-search-box.vrt.ts
```diff
@@ -0,0 +1,287 @@
...
+			<ontario-search-box caption="Search"></ontario-search-box>
+		`);
+		await page.waitForChanges();
+
+		const host = page.locator('ontario-search-box');
+		await expect(host).toHaveClass('hydrated');
+		await expect(host).toHaveScreenshot({
+			animations: 'disabled',
+			caret: 'hide',
+		});
```
smorrisods: I saw somewhere where there was a utility function for this, let's use that where we call this since it normalizes the `animations` and the `caret` properties.

smorrisods (changes requested):

packages/ontario-design-system-component-library/src/components/ontario-search-box/test/ontario-search-box.vrt.ts
muntalee: Good point! Found it and used it in aa1dfd5e

smorrisods (approved):
LGTM 🚀

PR #322, opened 2026-07-27
DS-2682: Implement autocomplete functionality in `ontario-search-box` component

muntalee:
## NOTICE

This PR was originally from #302, but was unfortunately lost due to technical difficulties.

## Summary
Adds complete autocomplete support to the `ontario-search-box` component with keyboard navigation, accessibility semantics, and flexible suggestion authoring modes (async suggestions, slotted `ontario-search-result-item`, and custom slotted HTML).

After the review comments, this was reworked to a segment-based highlight model so highlight behaviour is consistent across render paths.

## Merge Notice
This PR is dependent on DS-2703 (`ontario-search-result-item` component): https://github.com/ongov/ontario-design-system/pull/303

## Changes
- Enhanced `ontario-search-box.tsx`:
  - Props for autocomplete behaviour: `enableAutocomplete`, `getSuggestions`, `minChars`, `debounceMs`, `maxSuggestions`
  - Suggestion list state management (open/close/active/hover)
  - Keyboard navigation (Arrow keys, Enter, Escape, Tab)
  - ARIA combobox/listbox semantics (`aria-expanded`, `aria-controls`, `aria-activedescendant`, `aria-live`)
  - Selection handling for keyboard and mouse
  - Events: `autocompleteQueryUpdated`, `autocompleteSuggestionsUpdated`, `autocompleteSuggestionSelected`
- Introduced shared highlight model:
  - Added `Segment` type and `computeHighlightSegments` utility
  - Unified suggestion highlight behaviour (contiguous-first, fuzzy fallback)
- Updated `ontario-search-result-item`:
  - Added `segments` prop for highlight rendering
  - Kept legacy highlight fallbacks for compatibility
- Updated styling in `ontario-search-box.scss` to preserve visual behaviour and address review feedback
- Expanded test coverage:
  - Condition-based e2e checks for rendered autocomplete UI
  - Unit coverage for keyboard behaviour and highlight rendering paths

- Async suggestions via `getSuggestions(query)`
- Slotted `ontario-search-result-item` suggestions
- Slotted custom HTML suggestions


packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.scss
```diff
@@ -136,6 +142,127 @@ input[type='reset'].ontario-search__reset {
...
+	&:focus {
+		outline: none;
+	}
+}
+
+.ontario-search-autocomplete__suggestion-option:last-child {
+	padding-bottom: spacing.$spacing-2;
+}
+
+.ontario-search-autocomplete__suggestion-option--active {
```
erincarrieresontario: Can we combine this line with the one underneath since they have the same style rules?

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.scss
```diff
@@ -136,6 +142,127 @@ input[type='reset'].ontario-search__reset {
...
+	background-color: colours.$ontario-colour-white;
+	border-radius: 0 0 globalVariables.$global-radius globalVariables.$global-radius;
+	border-top: globalFunctions.px-to-rem(2) solid $ontario-search-autosuggest-box-shadow;
+	box-shadow: 0 globalFunctions.px-to-rem(3) globalFunctions.px-to-rem(8) globalFunctions.px-to-rem(1)
+		$ontario-search-autosuggest-box-shadow;
+	clip-path: inset(0 globalFunctions.px-to-rem(-8) globalFunctions.px-to-rem(-8) globalFunctions.px-to-rem(-8));
+	display: none;
+	list-style: none;
+	margin: globalFunctions.px-to-rem(-2) spacing.$spacing-0 spacing.$spacing-0 spacing.$spacing-0;
+	padding: 0;
```
erincarrieresontario: lets use a spacing variable for all the zero spacing values in this css

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.scss
```diff
@@ -136,6 +142,127 @@ input[type='reset'].ontario-search__reset {
...
+	display: none !important;
+}
+
+::slotted(.ontario-search-autocomplete__slot-option--active),
+::slotted(.ontario-search-autocomplete__slot-option--hovered) {
+	background-color: colours.$ontario-colour-button-tertiary--hover;
+}
+
+::slotted(.ontario-search-autocomplete__slot-option[aria-disabled='true']) {
+	opacity: 0.65;
```
erincarrieresontario: this opacity is re-used a few times, lets make it a variable at the top of the file

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +392,547 @@ export class OntarioSearchBox {
...
+	 * Handles autocomplete query updates for slotted and async suggestion modes.
+	 */
+	private async handleAutocompleteInput(query: string) {
+		this.autocompleteQueryUpdated.emit({ query });
+
+		if (!this.enableAutocomplete) return;
+
+		if (query.length < (this.minChars ?? OntarioSearchBox.DEFAULT_MIN_CHARS)) {
+			this.updateSlotSuggestionVisibility('');
+			this.closeSuggestions();
```
erincarrieresontario: We have these three lines in a few places, could we maybe throw this into a function called 'clearSuggestions' or something?

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +392,547 @@ export class OntarioSearchBox {
...
+	private getSuggestionLabel(index: number): string {
+		if (this.hasSuggestionSlotContent) {
+			const option = this.getSlotSuggestionElements()[index];
+			return this.getSlotOptionLabel(option);
+		}
+
+		return this.suggestions[index]?.label || '';
+	}
+
+	private handleSuggestionMouseOver = (event: MouseEvent) => {
```
erincarrieresontario, line 708: we should also probably handle mouseLeave

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.scss
muntalee: Merged `--active` and `--hovered` in 79f9b18 into a single selector since they share identical styles.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.scss
muntalee: Woops good idea. I replaced the bare `0` padding values in `.ontario-search-autocomplete__suggestion-list` and `.ontario-search__visually-hidden` with `spacing.$spacing-0` in 79f9b18

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.scss
muntalee: Good idea! I extracted the value into `$ontario-search-autocomplete-disabled-opacity` at the top of the file and referenced it in both the slotted and non-slotted disabled selectors in 79f9b18.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Yeah good idea! I extracted the repeated pattern into a `clearSuggestions()` private method and replaced all three call sites with it in 5dd29bdf.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: I added a `handleSuggestionMouseLeave` handler that resets the hovered index and re-decorates slot options, wired to `onMouseLeave` on the suggestion list in 5dd29bdf.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -276,6 +392,561 @@ export class OntarioSearchBox {
...
+			const parsedIndex = Number(fallbackIndex);
+			if (!Number.isNaN(parsedIndex)) {
+				return parsedIndex;
+			}
+		}
+
+		return -1;
+	}
+
+	private handleSuggestionMouseDown = (event: MouseEvent) => {
```
smorrisods, line 874: Tracing this through: `event.preventDefault()` here is doing two jobs at once. It suppresses the browser's default `mousedown` behaviour of shifting/blurring focus away from the input (the standard trick every combobox/listbox implementation uses to keep the input focused when clicking an option), *and* it commits the selection immediately via `this.selectSuggestionByIndex(...)` — on the down-event, not the up-event. Only the first part actually needs to happen on `mousedown`. The e2e test proves the second part is really happening: it dispatches a bare `mousedown` with no follow-up `click`/`mouseup` and asserts the value is already set and the list already closed — the down-event is the point of no return here, which is exactly what [WCAG 2.5.2 Pointer Cancellation](https://www.wcag.com/developers/2-5-2-pointer-cancellation/) exists to prevent (a user pressing down, then dragging off before releasing, should be able to cancel).

Let's keep `event.preventDefault()` on `mousedown` (still needed, still correct) but move `selectSuggestionByIndex(...)` to an `onClick` handler on the same element instead:
```tsx
private handleSuggestionMouseDown = (event: MouseEvent) => {
	if (!this.enableAutocomplete) return;
	if (!this.hasSuggestionSlotContent) return;
	event.preventDefault();
};

private handleSuggestionClick = (event: MouseEvent) => {
	if (!this.enableAutocomplete) return;
	const selectedIndex = this.getSuggestionIndexFromEvent(event);
	if (selectedIndex < 0) return;
	this.selectSuggestionByIndex(selectedIndex, 'mouse');
};
```
That gives us "up-event commits, down-event can abort" for free, satisfying 2.5.2's "Up Reversal" allowance, and it shouldn't require touching `BLUR_CLOSE_DELAY_MS` — that timer can stay as a fallback for slotted custom content that doesn't call `preventDefault()` itself.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -153,45 +209,75 @@ export class OntarioSearchBox {
...
 	@State() hintTextId: string | null | undefined;
+	@State()
+	private suggestions: AutocompleteSuggestion[] = [];
+	@State() private activeSuggestionIndex: number = -1;
+	@State() private hoveredSuggestionIndex: number = -1;
+	@State() private suggestionsOpen = false;
+	@State() private hasSuggestionSlotContent = false;
+	@State() private ariaLiveMessage = '';
 
 	@State() translations: any = translations;
```
smorrisods, line 252: We've got `@State() translations: any = translations;` again here — same shape as the issue on #256. Since `translations` is imported from the JSON i18n file, we can type this as `typeof translations` instead of `any`, and it'd be worth a short JSDoc on the state property while we're there.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -70,6 +99,33 @@ export class OntarioSearchBox {
...
+	 * Async suggestion provider for autocomplete mode.
+	 * Slot content has precedence over this callback.
+	 */
+	@Prop()
+	getSuggestions?: (query: string) => Promise<Suggestion[]>;
+
+	/**
+	 * Minimum number of characters required before suggestions are shown.
+	 */
+	@Prop() minChars?: number = 1;
```
smorrisods: The `@Prop` defaults for `minChars`, `debounceMs`, and `maxSuggestions` are hard-coded (`1`, `150`, `8`), but we also declare `DEFAULT_MIN_CHARS`, `DEFAULT_DEBOUNCE_MS`, and `DEFAULT_MAX_SUGGESTIONS` constants that are used elsewhere (e.g. lines 403/444/459/650) to fall back to the same values. Two sources of truth for the same defaults means they can drift if one is changed and not the other. Let's have the `@Prop` defaults reference the constants directly:
```ts
@Prop() minChars: number = OntarioSearchBox.DEFAULT_MIN_CHARS;
```

packages/ontario-design-system-component-library/src/utils/components/search-box-autocomplete.ts
```diff
@@ -0,0 +1,136 @@
+export interface Segment {
```
smorrisods, line 2: `resolveSuggestionSegments` has a JSDoc, but the exported `Segment` and `SegmentResolutionInput` interfaces, `computeHighlightSegments`, and the private `buildSegments` don't. Let's add JSDocs to all of these — even the private helper — since this file is going to be a shared util other components lean on.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.scss
```diff
@@ -136,6 +143,124 @@ input[type='reset'].ontario-search__reset {
...
+::slotted(.ontario-search-autocomplete__slot-option--hovered) {
+	background-color: colours.$ontario-colour-button-tertiary--hover;
+}
+
+::slotted(.ontario-search-autocomplete__slot-option[aria-disabled='true']) {
+	opacity: $ontario-search-autocomplete-disabled-opacity;
+	cursor: default;
+}
+
+.ontario-search-autocomplete__suggestion-description {
```
smorrisods, line 250: This hard-codes `line-height: 1.35`. Let's name it as a local scss variable here (e.g. `$search-autocomplete-suggestion-description-line-height`) instead of a raw number, so it reads intentionally and is easy to find if it ever needs to change.

packages/ontario-design-system-component-library/src/components/ontario-task/readme.md
```diff
@@ -146,7 +146,7 @@ Example of a task with a hint.
 
 The Ontario Task component supports server-side rendering, with a few considerations:
 
-- **Task status validation:** Always pass a valid `taskStatus`. Invalid values default to `'notStarted'` at runtime.
+- **Task status validation:** Always pass a valid `taskStatus`. Invalid values default to `'NotStarted'` at runtime.
```
smorrisods: This file is unrelated to search-box autocomplete, but this PR reverts a fix that landed in the base of this stack (#303, commit `c8589659`, "fix(ontario-task): correct stale taskStatus documentation to match real values"). It changes the documented values back from `notStarted` / `inProgress` to `NotStarted` / `InProgress`, which no longer match the real `TaskStatus` enum (`task-statuses.enum.ts`). Looks like an accidental revert from a rebase/merge — let's restore the lowercase values here.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -15,9 +15,34 @@ import {
...
+export interface AutocompleteSuggestion {
+	id?: string;
+	label: string;
+	value?: string;
+	description?: string;
+	href?: string;
+	disabled?: boolean;
+	segments?: Segment[];
+	/** @deprecated Use `segments` instead. */
+	boldRanges?: Array<{ start: number; end: number }>;
```
smorrisods: These are marked `` `@deprecated Use `segments` instead` `` (here, and mirrored on `ontario-search-result-item.tsx`'s `highlightParts`/`boldRanges` props), but `boldRanges`/`highlightParts` were only ever pre-release/in-progress shapes — they haven't shipped in a published version yet. There's no consumer to deprecate this for. Since we're not bound to backward compatibility here, let's just remove them outright rather than carry a deprecation notice for an API that never existed publicly — keeps the surface at the single `segments` field the design intended, instead of three fields (one live, two "deprecated").

smorrisods (changes requested):
Picking this back up — the autocomplete behaviour itself is solid and well tested. A few things worth addressing before merge: the pointer-cancellation issue on suggestion selection (committing on `mousedown` instead of `click` — see the concrete fix below, which doesn't require touching the blur-delay timer), the `any` type back on `translations`, duplicated default values that can drift from their named constants, some missing JSDocs on the new autocomplete util, a small named-vs-magic-number gap in the scss, a premature `@deprecated` on props that haven't shipped yet, and — the one that surprised me — this PR reverts a docs fix that just landed in #303 for `ontario-task`'s readme, which is worth restoring since it's unrelated to autocomplete.

---
🤖 A robot wrote this in Scott's voice. It does not, in fact, have opinions about `const`s — Scott does.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: So I split it into two handlers: `handleSuggestionMouseDown` only calls `event.preventDefault()` now to keep focus on the input, and a new `handleSuggestionClick` commits the selection on the up-event. Committed in 65105bee.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Tried `typeof` translations in 2785edb1 but it leaks the JSON path into the emitted `.d.ts` and breaks the Angular build, so I had to revert it in 3fa93248. Every other component in the repo uses any for the same reason; proper typing would need a cross-component follow-up.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Done in 03077949

packages/ontario-design-system-component-library/src/utils/components/search-box-autocomplete.ts
muntalee: Done in a7050e0a

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.scss
muntalee: Done in 3661e462

packages/ontario-design-system-component-library/src/components/ontario-task/readme.md
muntalee: Restored in 97959f74

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Removed in 678f7973

erincarrieresontario (approved):
nice changes @muntalee 👍

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
```diff
@@ -38,9 +39,21 @@ export class OntarioSearchResultItem {
...
+	 * Optional highlight parts used to style input-matched text and completion text.
+	 * Deprecated in favour of `segments`.
+	 */
+	@Prop() highlightParts?: Array<{ text: string; isInputMatch: boolean }>;
+
+	/**
+	 * Optional bold ranges over the label string for completion emphasis.
+	 * Deprecated in favour of `segments`.
+	 */
+	@Prop() boldRanges?: Array<{ start: number; end: number }>;
```
smorrisods: Can we please remove these properties as they've been superseded by segments.

packages/ontario-design-system-component-library/src/index.ts
```diff
@@ -1,2 +1,3 @@
 // export { Components, JSX } from './components';
 export * from './components';
+export * from './utils/components/search-box-autocomplete';
```
smorrisods: I'm going to recommend we remove this and make sure the export happens in our components or via the utilities rather than at the top level of the library.

smorrisods (changes requested):
I've noticed two remaining inconsistencies.

packages/ontario-design-system-component-library/src/components/ontario-search-result-item/ontario-search-result-item.tsx
muntalee: Addressed in dcddae16, I removed the superseded properties from `ontario-search-result-item` and now only segments is exposed for that API path.

packages/ontario-design-system-component-library/src/index.ts
muntalee: Done in dcddae16, I removed the top-level export from `src/index.ts` so the utility is no longer exported at the library root.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
```diff
@@ -15,9 +15,30 @@ import {
...
+}
+
+export type Suggestion = string | AutocompleteSuggestion;
+
+/** Payload emitted when a suggestion is selected */
+export interface AutocompleteSuggestionSelectedEvent {
+	query: string;
+	suggestion: AutocompleteSuggestion;
+	source: 'keyboard' | 'mouse';
+}
```
smorrisods: Our other components put these interfaces in another file, check them out for the pattern.

smorrisods (changes requested):
One minor knitpick on where the component interfaces are placed.

packages/ontario-design-system-component-library/src/components/ontario-search-box/ontario-search-box.tsx
muntalee: Good idea! I moved the interface and adjusted the code here 954a5ec.

smorrisods (approved):
LGTM 🚀

PR #42, opened 2025-04-16
DS-2186: Fixes Angular PoC errors

MeaghanCarrieres:
This MR updates the SLEW of errors that were appearing when running the app-angular PoC. This was done by updating the `app-angular` project to ensure compatibility with the latest version of the `ontario-design-system-component-library-angular` package. The following changes were made:

### Key Updates

**Angular and Typescript version updates**
- Updated Angular dependencies to version `^18.2.11` to match the requirements of the `ontario-design-system-component-library-angular` package.
- Updated TypeScript to the compatible version required by Angular `^18.2.11`.

**Custom Element Schema**
Added `CUSTOM_ELEMENTS_SCHEMA` to the NgModule to support the use of custom elements from the `ontario-design-system-component-library-angular`. (This is actually outlined in our debugging FAQs in the angular component library docs)

### Testing
- Verified that the application builds and serves correctly using `pnpm run start`.
- Tested the integration of components from the `ontario-design-system-component-library-angular` to ensure they render without schema-related errors.
- Confirmed no breaking changes in the application after the updates.

PR #323, opened 2026-07-27
Fix semantic-release/github crash from bogus Renovate issue references

smorrisods:
## Summary

Fixes the `Publish Packages` job failure seen on [run 30281151214](https://github.com/ongov/ontario-design-system/actions/runs/30281151214/job/90029869964) (the `chore(release): 9.0.0` release), where the job failed with:

```
Error: Could not resolve to an issue or pull request with the number of 8203
```

despite npm publish and the GitHub release already succeeding. This was cosmetic-but-blocking: it failed CI and skipped downstream jobs (`Sync Docs to Docusaurus`, `Enable Docs Auto-Merge`).

## Root cause chain

1. Renovate always HTML-entity-escapes scoped package names in PR bodies/commit messages as the literal string `&#8203;` (e.g. `@&#8203;ngx-translate/http-loader`), to stop GitHub from auto-pinging a nonexistent `@ngx-translate` user. That literal string always contains the substring `#8203`.
2. `conventional-commits-parser`'s per-line reference scanning (used transitively via `conventional-changelog-angular`, which `@semantic-release/commit-analyzer`, `@semantic-release/release-notes-generator`, and `semantic-release-lerna`'s own notes generator all depend on) records **any** bare `#<digits>` as a reference — tagged `action: null` — whenever a line doesn't contain an actual close/fix/resolve keyword immediately adjacent. This is by-design fallback behaviour in the parser (needed elsewhere for passive "see #123" style mentions).
3. `conventional-changelog-angular`'s `commit.hbs` template renders `, closes #N` for **every** entry in `references`, without checking whether `action` was ever set — so the `action: null` entry for `#8203` still got rendered as `, closes #8203` in the generated release notes/changelog.
4. `@semantic-release/git`'s commit message template (`chore(release): ${nextRelease.version}\n\n${nextRelease.notes}`) bakes that polluted notes text verbatim into the real `chore(release): X.Y.Z` commit.
5. On the *next* release run, `issue-parser` (used by `@semantic-release/github`, which has a stricter, correct keyword-adjacency regex) now finds a legitimate-looking `closes #8203` in that prior release commit's own message, and passes `{ number: 8203 }` into `@semantic-release/github`'s `success.js`, which has **no error handling** around its GraphQL/REST lookups — so GitHub's "could not resolve" error crashes the whole `success` step and the job.

This already happened once silently on `9.0.0-alpha.1` (June 25) before recurring on the real `v9.0.0` release, and is self-perpetuating on every Renovate PR for a scoped package.

Confirmed as a real, unfixed upstream bug: [semantic-release/github#1092](https://github.com/semantic-release/github/issues/1092) (open, filed Aug 2025, same failure class — an unresolvable issue/PR reference crashing an already-successful `success` step).

## Fix

Two `pnpm patch`es, following the repo's existing precedent (`patches/@semantic-release__github@12.0.2.patch`, removed in `ae508eae` once a related bug in the same file was fixed upstream in v12.0.3):

1. **`patches/@semantic-release__github@12.0.9.patch`** — wraps the three unguarded network calls in `lib/success.js` (associated-PRs GraphQL query, PR-verification `pFilter` block, related-issues GraphQL query) in `try/catch`. A single unresolvable reference now logs a warning and is skipped instead of crashing the whole job. (Defense in depth / resilience fix.)
2. **`patches/conventional-changelog-angular@8.0.0.patch`** — fixes `src/templates/commit.hbs` to only render `, closes #N` when the reference's `action` is a real close/fix/resolve keyword, not for every parsed reference. Stops the bogus text from ever being generated in the first place. (Root-cause fix.)

## Verification

- Extracted the real Renovate commit (`036fd71`) and release commit (`a565767`) text from `master` history and ran them through the actual installed `conventional-commits-parser` / `conventional-changelog-writer` / patched `commit.hbs` — confirmed the `#8203` reference (`action: null`) no longer renders `, closes #8203`, while a real `closes #42`-style reference still renders correctly.
- `node --check` on both patched files.
- `pnpm install --frozen-lockfile` succeeds cleanly after `patch-commit`.

## Follow-up (not addressed here)

The already-poisoned `chore(release): 9.0.0` commit/tag on `master` contains the literal "closes #8203" text, but this is self-limiting: future semantic-release runs only analyze commits *after* the last release tag, so it won't be re-scanned. No cleanup action is planned unless it resurfaces.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,5 +70,5 @@
 		"@nrwl/nx-win32-x64-msvc": "15.9.3",
 		"bl": "^7.0.0"
 	},
-	"packageManager": "pnpm@11.10.0+sha512.0b7f8b98060031904c017e3a41eb187a16d40eeb829b95c4f8cb03681761fc4ab53dd219115b9b447f4dce1a05a214764461e7d3703392a9f32f9511ce8c86c8"
+	"packageManager": "pnpm@11.10.0"
 }
--- a/package.json
+++ b/package.json
@@ -70,5 +70,5 @@
 		"@nrwl/nx-win32-x64-msvc": "15.9.3",
 		"bl": "^7.0.0"
 	},
-	"packageManager": "pnpm@11.10.0"
+	"packageManager": "pnpm@11.10.0+sha512.0b7f8b98060031904c017e3a41eb187a16d40eeb829b95c4f8cb03681761fc4ab53dd219115b9b447f4dce1a05a214764461e7d3703392a9f32f9511ce8c86c8"
 }
--- a/patches/@semantic-release__github@12.0.9.patch
+++ b/patches/@semantic-release__github@12.0.9.patch
@@ -1,0 +1,117 @@
+diff --git a/lib/success.js b/lib/success.js
+index 48246f29a2162ce0abc0d044112de94bab39becb..e106ddf9ba078c9f75ef06ef1c358183654af40f 100644
+--- a/lib/success.js
++++ b/lib/success.js
+@@ -78,10 +78,20 @@ export default async function success(pluginConfig, context, { Octokit }) {
+     // Get associatedPRs
+     const associatedPRs = await inChunks(shas, 100, async (chunk) => {
+       const responsePRs = [];
+-      const { repository } = await octokit.graphql(
+-        buildAssociatedPRsQuery(chunk),
+-        { owner, repo },
+-      );
++      let repository;
++      try {
++        ({ repository } = await octokit.graphql(
++          buildAssociatedPRsQuery(chunk),
++          { owner, repo },
++        ));
++      } catch (error) {
++        logger.error(
++          "Failed to resolve associated pull requests for commits %O: %s",
++          chunk,
++          error.message,
++        );
++        return responsePRs;
++      }
+       const responseAssociatedPRs = Object.values(repository).map(
+         (item) => item.associatedPullRequests,
+       );
+@@ -118,26 +128,35 @@ export default async function success(pluginConfig, context, { Octokit }) {
+     const uniqueAssociatedPRs = uniqBy(flatten(associatedPRs), "number");
+ 
+     const prs = await pFilter(uniqueAssociatedPRs, async ({ number }) => {
+-      const commits = await octokit.paginate(
+-        "GET /repos/{owner}/{repo}/pulls/{pull_number}/commits",
+-        {
+-          owner,
+-          repo,
+-          pull_number: number,
+-        },
+-      );
+-      const matchingCommit = commits.find(({ sha }) => shas.includes(sha));
+-      if (matchingCommit) return matchingCommit;
+-
+-      const { data: pullRequest } = await octokit.request(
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -70,5 +70,5 @@
 		"@nrwl/nx-win32-x64-msvc": "15.9.3",
 		"bl": "^7.0.0"
 	},
-	"packageManager": "pnpm@11.10.0"
+	"packageManager": "pnpm@11.10.0+sha512.0b7f8b98060031904c017e3a41eb187a16d40eeb829b95c4f8cb03681761fc4ab53dd219115b9b447f4dce1a05a214764461e7d3703392a9f32f9511ce8c86c8"
```
smorrisods: Let's not add in the sha

smorrisods: Good catch, that hash got auto-appended by `corepack use`. Reverted to plain `pnpm@11.10.0` (matching develop) in ca917d35.

muntalee (approved):
LGTM 🐐

patches/@semantic-release__github@12.0.9.patch
```diff
@@ -0,0 +1,117 @@
...
++        return shas.includes(pullRequest.merge_commit_sha);
++      } catch (error) {
++        logger.error(
++          "Failed to verify pull request #%d, skipping it: %s",
++          number,
++          error.message,
++        );
++        return false;
++      }
+     });
```
erincarrieresontario, line 83: Could this result in releases succeeding with incomplete metadata if GitHub is having issues? It might be worth emitting a warning that release notes may be incomplete when we hit these error paths.

erincarrieresontario (commented):
Looks like there are some merge conflicts @smorrisods , otherwise looks good! Just one comment/question!

patches/@semantic-release__github@12.0.9.patch
smorrisods: Good question. Tracing it through, rate limits actually take a different path — `lib/octokit.js` configures the `throttling` plugin with `onRateLimit`/`onSecondaryRateLimit`, which auto-retries the request (up to `RETRY_CONF.retries`) and already logs its own `octokit.log.warn` before we'd ever hit these `catch` blocks.

These catches are for the harder failure (bogus `closes #N` refs breaking the whole GraphQL query), and each one already calls `logger.error` with the chunk/PR/issue numbers. It's also worth noting `success.js` runs after `publish` has already created the release/tag/notes — the impact here is scoped to skipping the "released" comment/label on the affected issues/PRs, not the release notes themselves.

Happy to add a rollup warning at the end if we want it more visible, but wanted to confirm the actual blast radius first.

---
🤖 This comment was summoned by CoPilot doing a Scott cosplay.

erincarrieresontario: I think thats fine then - approved!

erincarrieresontario: oop, actually @smorrisods there are some merge conflicts

andrewpeart (approved):

PR #317, opened 2026-07-24
Alpha -> Production

smorrisods:
## Summary

Promotes `alpha` → `master` for the next production release. `alpha` has been validated and is approved for release; this PR brings its full contents into production.

**Scope**: 196 commits, 157 files changed (+11,446 / -8,278), spanning `ontario-design-system-component-library`, `app-nextjs`, `app-angular`/`app-react`, the React/Angular/HTML-generator wrappers, and CI/build tooling.

## ⚠️ Breaking change

- **`ontario-card`** (DS-2258): multi-word `headerColour` prop values are now camelCase (e.g. `dark-accent` → `darkAccent`, `light-teal` → `lightTeal`) to match `ontario-badge`'s convention. Single-word values (`sky`, `gold`, etc.) are unaffected. Rendered CSS class names are unchanged.

## Highlights

**Accessibility & component fixes**
- `ontario-icon`: decorative icons are now correctly non-focusable (DS-2246)
- `ontario-card`: fixed duplicate/missing tab stops and background gaps on uneven card heights (DS-2625, DS-2634)
- `ontario-button`: fixed `htmlType` submit/reset behaviour across the shadow DOM boundary (DS-1762)
- `ontario-input`: form value now syncs correctly when the `value` prop changes externally
- `ontario-table`: fixed SSR highlight-prop breakage in Next.js (DS-2236), moved highlight indicator fill to a design token, inlined its SVG for portability
- `ontario-header`: accessibility updates (DS-2597)
- Checkbox/radio controls: added white background fill (DS-2652)
- New `ontario-summary-list` / `ontario-summary-list-item` components implemented (feature work landing as part of this promotion)

**Testing**
- Migrated `ontario-aside` tests from Puppeteer to Playwright, added new E2E coverage (DS-2580)
- Added/expanded unit and E2E coverage for `ontario-card`, `ontario-button`, `ontario-input`, `ontario-table`
- Regenerated Webkit VRT baselines for the Playwright 1.60 upgrade

**Build & dependency upgrades**
- TypeScript 6 compatibility across the monorepo (tsconfig updates, `moduleResolution`, `ignoreDeprecations`)
- Angular upgraded to 22.0.x; Node engine requirement bumped to 22.22.3
- `@playwright/test` → 1.60.0, `@stencil/playwright` → 0.4.3, `next` → 15.5.14 (security), `@ngx-translate/*` → v17
- Local Docker-based Playwright runners reorganized (UID/GID sync, arg forwarding, CI container alignment)

[...]

PR #281, opened 2026-06-23
Develop -> Alpha

smorrisods:
## Summary

Promotes all changes accumulated on `develop` into `alpha` in preparation for the **v9.0.0** release.

`master` is currently at **v8.1.0** (stable). This PR brings the development branch forward and includes at least one breaking change that will trigger a major version bump to **9.0.0** when released.

---

### `ontario-card` — `headerColour` values are now camelCase (DS-2258)

Multi-word `headerColour` prop values have been renamed from kebab-case to camelCase. Single-word values (`sky`, `gold`, etc.) are unaffected.

| Old value (kebab-case) | New value (camelCase) |
|---|---|
| `dark-accent` | `darkAccent` |
| `light-accent` | `lightAccent` |
| `light-gold` | `lightGold` |
| `light-yellow` | `lightYellow` |
| `light-taupe` | `lightTaupe` |
| `light-green` | `lightGreen` |
| `light-lime` | `lightLime` |
| `light-teal` | `lightTeal` |
| `light-sky` | `lightSky` |
| `light-blue` | `lightBlue` |
| `light-purple` | `lightPurple` |
| `light-orange` | `lightOrange` |
| `light-red` | `lightRed` |
| `light-magenta` | `lightMagenta` |

**Migration:** Update any `header-colour` attribute or `headerColour` prop values in templates and code to use the new camelCase equivalents.

### `ontario-header` — `menuButtonToggled` event payload changed (DS-2597)

The `menuButtonToggled` custom event now emits a `HeaderMenuToggleDetail` object instead of a plain `boolean`.

Before:
```ts
// event.detail was a boolean
element.addEventListener('menuButtonToggled', (e) => {
  const isOpen: boolean = e.detail;
});
```

After:
```ts
// event.detail is now HeaderMenuToggleDetail
interface HeaderMenuToggleDetail {
  isOpen: boolean;
  trigger: 'click' | 'keyboard' | 'programmatic';
}
element.addEventListener('menuButtonToggled', (e) => {
  const { isOpen, trigger } = e.detail;
});
```

**Migration:** Update any listener that reads `event.detail` directly as a boolean to instead read `event.detail.isOpen`.

---

## New Components

- **`ontario-summary-list`** (DS-2655) — Groups labelled answers for review before a user submits or confirms information. Supports `caption`, `headingLevel`, `fullWidth`, `columnRatio`, `captionActionLink`, and `language` props, plus a `caption-action` slot for router-aware links.
[...]

PR #324, opened 2026-07-27
Production -> Develop

smorrisods:
## Production -> Develop

Merges `master` (currently `v9.0.0`) back into `develop`, following the standard `Alpha -> Production` / `Production -> Develop` promotion pattern used elsewhere in this repo's history (e.g. #281, #286, #317).

This is a manual merge via an intermediate branch (`merge/master-into-develop-v9.0.0`, off `develop`, with `master` merged in) rather than a direct PR from `master`, so the merge conflicts below could be reviewed/resolved and the full test suites re-verified locally before opening the PR.

### Conflicts resolved

- **`.github/workflows/sync-docs-to-docusaurus.yml`**: `develop` still had the old nested-checkout-under-`packages/docusaurus` layout; `master` already includes #316's fix (`refactor(ci): stop nesting the Docusaurus checkout inside packages/*`), which removed the monorepo checkout, un-nested all paths, and dropped the now-moot `Validate docs-only change scope` step. Kept master's un-nested structure (the real, documented fix) but preserved `develop`'s newer `actions/checkout@v7` pin (a Renovate bump `master` hasn't received yet, since that bump only landed on `develop`).
- **`packages/ontario-design-system-component-library/src/components.d.ts`**: autogenerated Stencil output (excluded from Prettier per `.prettierignore`). Rather than hand-merge the raw diff, regenerated via `stencil build --docs` against the merged component sources so it's exactly what Stencil would emit — ended up textually identical to `develop`'s existing version, since the two branches' actual component sets had already converged through prior promotions.

### Verification

- `pnpm run build-libs` — all 7 library projects build clean.
- `pnpm run build-apps` — all demo apps build clean.
- `pnpm run test:unit` — 225 passed, 2 skipped.
- `pnpm run test:e2e` — component library: 70/70 passed (chromium); `app-nextjs`: 27/27 passed (chromium/firefox/webkit).
- Confirmed no leftover conflict markers and no unexpected file changes outside the two conflicted files + routine version/lockfile bumps.

Once merged, `develop`'s version fields will match the real released `v9.0.0` (previously stuck at a stale `9.0.0-alpha.1`).


erincarrieresontario (approved):
LGTM 👍

PR #330, opened 2026-07-30
DS-912: Allow options to be passed into dropdown

erincarrieresontario:
- Cleaned up how options get into the dropdown list component. Instead of letting people jam native <option> tags into the light DOM (which never actually worked properly since browsers won't slot options into a shadow DOM select anyway), we built a proper ontario-dropdown-option component to match the pattern we already use elsewhere (like task-list/task). 
- You can still pass options in via the options prop as before, but now if you want to write them out as markup, you use <ontario-dropdown-option> instead, and it just gets parsed and turned into real options under the hood. 
- Added tests for it, updated the readme, and got everything rebuilt so the React/Angular wrappers and type definitions are all in sync.


packages/ontario-design-system-component-library/src/components/ontario-dropdown-list/readme.md
```diff
@@ -618,6 +618,24 @@ For more guidance, visit the [Error messaging guidance page](https://designsyste
...
+	<option value="dropdown-option-3">Option 3</option>
+</ontario-dropdown-list>
+```
+
+Notes:
+
+- These `<option>` children are read once from the host's light DOM and re-rendered inside the component's shadow root; they are not projected via a native `<slot>`. Browsers do not support slotting `<option>` elements into a `<select>` rendered in a shadow root, so a `<slot>`-based approach silently fails to display any options.
+- If both `options` and `<option>` children are provided, the `options` prop takes precedence and the `<option>` children are ignored.
+- Use the `isEmptyStartOption` prop to render an empty/placeholder start option rather than including an `<option value="">` child directly; a slotted `<option>` with an empty `value` will be ignored in favour of the `isEmptyStartOption` prop, to avoid rendering a duplicate blank option.
+
```
smorrisods: Since browsers don't like using regular options and we're processing them into the component I think we should consider adding our own options component to keep the container pattern consistent rather than trying to repurpose the browser's native elements?

smorrisods (changes requested):

packages/ontario-design-system-component-library/src/components/ontario-dropdown-list/readme.md
erincarrieresontario: yeah that's a good idea and keeps it in line with our other components

packages/ontario-design-system-component-library/src/components/ontario-dropdown-option/ontario-dropdown-option.tsx
```diff
@@ -0,0 +1,55 @@
...
+@Component({
+	tag: 'ontario-dropdown-option',
+	shadow: false,
+})
+export class OntarioDropdownOption {
+	/**
+	 * The value for the dropdown option.
+	 * Each value must be unique to the option within the parent `ontario-dropdown-list`.
+	 */
+	@Prop() value: string;
```
muntalee, line 37: Small nitpick: `value` is required here, but `ontario-dropdown-list.tsx`'s `getSlottedOptions()` falls back to the text content when `value` is missing, so it's actually optional at runtime. Should this be `value?: string;` to match?

muntalee (changes requested):

packages/ontario-design-system-component-library/src/components/ontario-dropdown-option/ontario-dropdown-option.tsx
```diff
@@ -0,0 +1,55 @@
...
+ * @example
+ * <ontario-dropdown-list caption="Do you like cats?" name="cat-dropdown">
+ *   <ontario-dropdown-option value="dropdown-option-1">Option 1</ontario-dropdown-option>
+ *   <ontario-dropdown-option value="dropdown-option-2" selected>Option 2</ontario-dropdown-option>
+ *   <ontario-dropdown-option value="dropdown-option-3">Option 3</ontario-dropdown-option>
+ * </ontario-dropdown-list>
+ */
+@Component({
+	tag: 'ontario-dropdown-option',
+	shadow: false,
```
smorrisods, line 30: I think this should be shadowed line all our other components.

packages/ontario-design-system-component-library/src/components/ontario-dropdown-list/ontario-dropdown-list.tsx
```diff
@@ -344,13 +363,56 @@ export class OntarioDropdownList implements Dropdown {
...
+	 * Reads `ontario-dropdown-option` elements passed as light DOM children of the host and
+	 * converts them into the same `DropdownOption[]` shape used by the `options` prop.
+	 *
+	 * An empty-value option (for example `<ontario-dropdown-option value="">Please select</ontario-dropdown-option>`)
+	 * is skipped; this placeholder pattern is instead handled through the `isEmptyStartOption` prop,
+	 * so that slotted markup can't produce a duplicate blank entry alongside the generated empty
+	 * start option.
+	 */
+	private getSlottedOptions(): DropdownOption[] {
+		const optionElements = Array.from(this.element.querySelectorAll<HTMLElement>(':scope > ontario-dropdown-option'));
```
smorrisods, line 386: `getSlottedOptions()` reads each child's data with `optionElement.getAttribute('value')` and `optionElement.hasAttribute('selected')` — i.e. it goes straight to the HTML attributes rather than the Stencil component's `.value`/`.selected` properties.

That works fine for plain HTML/Stencil markup, but tracing it through for the React/Angular wrappers this PR also regenerates: `createComponent` (react-output-target) and the Angular proxy both set props on the element as JS properties, not attributes. Since `value` and `selected` on `OntarioDropdownOption` aren't declared with `reflect: true`, setting them via the wrapper components updates the underlying prop but never touches the DOM attribute — so `getAttribute('value')`/`hasAttribute('selected')` would silently return the default/empty value for any consumer using `<OntarioDropdownOption>` from the React or Angular packages.

I think what you're looking for here is to read the properties directly instead of the attributes, since Stencil populates `.value`/`.selected` from the initial attribute on upgrade *and* keeps them in sync when set as a property:
```ts
const optionElements = Array.from(
  this.element.querySelectorAll<HTMLOntarioDropdownOptionElement>(':scope > ontario-dropdown-option'),
);
// ...
value: optionElement.value ?? optionElement.textContent?.trim() ?? '',
selected: !!optionElement.selected,
```
Worth double-checking against an actual React/Angular consumer test case — right now the e2e coverage only exercises the raw custom element via `page.setContent`, which wouldn't catch this.

packages/ontario-design-system-component-library/src/components/ontario-dropdown-list/ontario-dropdown-list.tsx
```diff
@@ -344,13 +363,56 @@ export class OntarioDropdownList implements Dropdown {
...
+		}
+
+		return optionElements
+			.filter(
+				(optionElement) => (optionElement.getAttribute('value') ?? optionElement.textContent?.trim() ?? '') !== '',
+			)
+			.map((optionElement) => ({
+				value: optionElement.getAttribute('value') ?? optionElement.textContent?.trim() ?? '',
+				label: optionElement.textContent?.trim() ?? '',
+				selected: optionElement.hasAttribute('selected'),
```
smorrisods, line 412: The fallback expression `optionElement.getAttribute('value') ?? optionElement.textContent?.trim() ?? ''` is repeated three times in this method (the empty-value check, the filter, and the map). Let's pull it into a small named helper, e.g. `getOptionValue(optionElement)`, so the "what counts as this option's value" logic only lives in one place — if we ever change the empty-value rule we'd otherwise have to remember to update it in three spots.

packages/ontario-design-system-component-library/src/components/ontario-dropdown-list/ontario-dropdown-list.tsx
```diff
@@ -344,13 +363,56 @@ export class OntarioDropdownList implements Dropdown {
 					.printMessage();
 				this.internalOptions = [];
 			}
+		} else {
+			// No `options` prop was provided; fall back to any slotted `ontario-dropdown-option` light DOM children.
+			this.internalOptions = this.getSlottedOptions();
 		}
```
smorrisods, line 369: `getSlottedOptions()` only runs once, from `parseOptions()` during `componentWillLoad`. Is that intentional — i.e. is dynamically adding/removing `<ontario-dropdown-option>` children after initial render explicitly out of scope? Worth a one-line callout in the readme/JSDoc if so, since it's a bit of a surprise coming from the native `<select>`/`<option>` mental model, where DOM mutations just work.

packages/ontario-design-system-component-library/src/components/ontario-dropdown-option/ontario-dropdown-option.tsx
```diff
@@ -0,0 +1,55 @@
+import { Component, Prop } from '@stencil/core';
+
+/**
+ * Ontario Dropdown Option represents a single, pre-defined option to be consumed by an
```
smorrisods, line 4: Nice pattern here mirroring the existing `task-list`/`task` light-DOM-holder approach, and the JSDoc placement/guidance links follow the established convention (`ontario-badge.tsx`, etc.) exactly. Good catch documenting *why* a `<slot>` won't work for this case too — that'll save someone a debugging session down the road.

smorrisods (changes requested):
Nice work here — this is a clean way to finally let people write options declaratively without relying on the (never-really-working) native `<option>` slotting trick. Docs and tests are thorough.

Found one correctness issue worth fixing before merge: `getSlottedOptions()` reads `value`/`selected` off the DOM attributes rather than the component properties, which breaks when `ontario-dropdown-option` is used through the regenerated React/Angular wrappers (they set properties, not attributes). Also flagged the `shadow: false` inconsistency, a small DRY nit, and a non-blocking question about dynamic children. Details inline.

---

packages/ontario-design-system-component-library/src/components/ontario-dropdown-option/ontario-dropdown-option.tsx
erincarrieresontario: Yes you are right - thanks @muntalee :)

packages/ontario-design-system-component-library/src/components/ontario-dropdown-list/ontario-dropdown-list.tsx
erincarrieresontario: okay! I fixed the slotted dropdown options to read their Stencil properties, ensuring React and Angular consumers preserve value and selected correctly.

packages/ontario-design-system-component-library/src/components/ontario-dropdown-list/ontario-dropdown-list.tsx
erincarrieresontario: okay I documented that slotted <ontario-dropdown-option> children are read only during initialization, later DOM additions or removals aren’t observed

packages/ontario-design-system-component-library/src/components/ontario-dropdown-list/ontario-dropdown-list.tsx
erincarrieresontario: Refactored the repeated option-value fallback into a getOptionValue() helper and updated all three call sites.

muntalee (approved):
LGTM @erincarrieresontario 🐐

PR #331, opened 2026-07-30
DS-2660:  added bottom offset for back-to-top

erincarrieresontario:
**What this does:**

This adds a way to control the position of the "Back to Top" button so it doesn't overlap with other floating buttons on the page (like a feedback button or live chat widget).

**The problem:**

The Back to Top button always sits a fixed distance from the bottom of the screen. If another floating button is also placed near the bottom of the page, the two can end up overlapping or stacking awkwardly, especially when the browser window is resized.

**The fix:**

Added a new optional setting called 'bottomOffset' that lets developers add extra spacing below the Back to Top button, so there's room for another button underneath it. For example, setting 'bottomOffset="63px"' pushes the button up by an extra 63 pixels from its normal position, without changing anything else about how it looks or behaves. If this setting isn't used, the button behaves exactly as it did before — nothing changes for existing pages.


packages/ontario-design-system-component-library/src/components/ontario-back-to-top/ontario-back-to-top.tsx
```diff
@@ -29,6 +29,22 @@ export class OntarioBackToTop {
...
+	 * launcher) are stacked below the Back to Top button, and space needs to be reserved so the two
+	 * don't overlap as the viewport is resized.
+	 *
+	 * The value is added on top of the existing `5%` offset (that is, `bottom: calc(5% + <bottomOffset>)`),
+	 * rather than replacing it.
+	 *
+	 * @example
+	 * <ontario-back-to-top bottom-offset="63px"></ontario-back-to-top>
+	 */
+	@Prop() bottomOffset?: string;
```
smorrisods, line 46: This is currently a freeform field with no guardrails directly adding CSS to the button inside the component.  

A few thoughts:
1. If we're going to use a property we should validate the value a user has passed us to avoid CSS injection into the component.
2. I think we should document how to do this with CSS vars rather than have a freeform field - with an option to turn off placement.
3. Perhaps we should rethink how this button is placed and find a solution that is easy to configure that is in line with other established patterns (eg. maybe the button doesn't even do its own placement at all and it's purely CSS)

smorrisods: Looked into how other design systems handle this. Here's what I found:

- **GOV.UK and USWDS don't ship a floating "back to top" component at all.** USWDS has no such component in its library, and GOV.UK doesn't either.
- GOV.UK's other "urgent floating-ish" components deliberately avoid `position: fixed`: the cookie banner and "Exit this page" button both render as plain in-flow elements that the consuming service places in its own markup wherever it wants — placement is pushed back to the consumer rather than owned by the component.
- **MUI's Floating Action Button** (the closest analog) ships with no built-in fixed positioning either — the docs show the consumer wrapping it in their own CSS with `position: fixed` themselves.

That said, it's not black-and-white: the button's whole purpose requires *some* fixed positioning to work by default, and CSS custom properties (the `--ontario-back-to-top-bottom-offset` var already used here) are the standard, sanctioned escape hatch for styling into shadow DOM — so exposing one isn't unusual on its own. The real problem is more the packaging: an unvalidated string `@Prop` bolted on top of it, rather than just documenting the CSS variable directly.

A few concrete alternatives, roughly in order of "how much the component still owns":

1. **Drop the `@Prop` entirely, keep only the CSS custom property.** Document `--ontario-back-to-top-bottom-offset` as a supported override consumers set via plain CSS on `ontario-back-to-top { ... }`. No unvalidated string flowing through Stencil/JS, no new versioned API surface.
2. **Expose a `::part()` on the button** instead, letting consumers override `position`/`bottom`/`right`/`z-index` fully via `::part(button) { ... }`. Maximum flexibility, no bespoke API to validate or maintain.
3. **Don't self-position at all** — render as a plain in-flow button and let the consumer's own layout decide if/how it's fixed or sticky (mirrors GOV.UK's approach). Cleanest conceptually, but loses the "drop it in, it just floats" zero-config ergonomics most teams probably want.
4. **Solve it design-system-wide, not per-component:** a shared token/custom-property convention (e.g. `--ontario-fixed-stack-offset`) that any current or future fixed/sticky component reads, set once by the app shell — scales past just one sibling widget without a new bespoke prop each time a new floating component ships.

Property validation is still required if it is to be kept in any form, regardless of which direction we go here.

---
🤖 Research assembled by your friendly neighbourhood CoPilot.

smorrisods (changes requested):
This is an interesting solution, I think we should discuss how we're positioning the button as we may be running into territory where we're now fighting our users and their needs. I've left some research in my comments based on a weird feeling I had reviewing this.

muntalee (approved):
LGTM 🐐 I'm wondering if this could eventually become part of a broader approach for floating/sticky elements across other components.

packages/ontario-design-system-component-library/src/components/ontario-back-to-top/ontario-back-to-top.tsx
erincarrieresontario: I updated the Back to Top component to make the bottomOffset prop safer and easier to use. It now validates that the value is a safe CSS length, ignores invalid input instead of applying it, and the docs/tests were updated to show the intended CSS variable-based usage. Check it out @smorrisods 👍

PR #337, opened 2026-08-04
Align in-page navigation link colour state and add visited-state VRT coverage

muntalee:
## Summary
This PR fixes the in-page navigation link colour behaviour and with that updated the VRT coverage for visited-link states. It updates the component styling for link states, adds explicit visited-state VRT cases for both in-page navigation components, and refreshes the related snapshot baselines.

## Changes
- Updated in-page navigation item link styling to align expected link colour behaviour.
- Added a new visited-state visual regression test case to the in-page navigation suite.
- Added a new visited-state visual regression test case to the in-page navigation item suite.
- Regenerated in-page navigation and in-page navigation item VRT snapshots after test updates.


erincarrieresontario (approved):
LGTM!

smorrisods (commented):
LGTM 🚀

smorrisods (approved):

PR #338, opened 2026-08-06
DS-2191: consistent assignment and display of checkbox-bg between shadow and real DOMs

andrewpeart:
### Summary
Fix for consistent `--checkbox-bg` variable assignment and display between real DOM and shadow DOM integrations.

::root is leveraged in the real DOM while ::host is used in the shadow DOM. They are mutually exclusive resulting in no collision.

### Why is this needed?
When applying the complete styles package plain HTML elements, :host targets are not in scope of the real DOM. Adding :root as an alternative selector allows for global assignment of variables, exclusively within context, while :host is ignored. And, vice-versa within the scope of a shadow DOM - :host is applied while :root is ignored.

### Tests
**Unit**
image

**E2E**
image

**Visual Verification**
Web component scope (:host)
| Before (:host only) | After (:root, :host) | Angular (:root, :host) | React (:root, :host) |
|--------|--------|--------|--------|
| image | image | image | 
image |


smorrisods:
Hey @andrewpeart, your branch is out of date with `develop` and won't be mergable until the new CI jobs are merged in.

smorrisods (approved):
LGTM 🚀

erincarrieresontario (approved):
👍

PR #342, opened 2026-08-07
Make semantic-release github success step non-fatal (hotfix)

smorrisods:
## Summary

Follow-up to #323, which fixed the `Publish Packages` crash from a bogus issue reference (Renovate's `&#8203;` escape misread as `#8203` by the commit parser) by wrapping specific GraphQL/REST calls in `success.js` in try/catch.

That fix worked, but its own commit body described the bug using a close-style keyword directly in front of that same bogus number - which the strict `issue-parser` in `success.js` reads as a real close-directive. So the exact same crash recurred on the next release ([9.1.0-alpha.1](https://github.com/ongov/ontario-design-system/actions/runs/31193231134/job/92918622714)), caused by the fix's own description rather than a new bug.

As before, npm publish and the GitHub release had **already succeeded** - only the post-release PR/issue commenting-and-labelling step failed, which skipped docs sync for that release.

## Fix

Two changes:

1. **Outer safety net.** Instead of guarding more individual call sites, wrap the whole `success()` step: renamed it to `successImpl` and added a `success` entry point that runs it and logs (doesn't rethrow) anything that escapes. This step is pure best-effort notification after the release is already live - it shouldn't be able to fail CI.
2. **Partial-batch recovery.** The existing per-batch GraphQL guards (up to 100 commits/issues per query) discarded the *whole batch* on a single bad alias, silently skipping every other legitimate PR/issue in it too. GitHub's GraphQL API actually returns partial data alongside the error (only the failing alias is nulled), so a new `graphqlWithPartialFallback` helper recovers that partial data instead of losing the whole batch - only the one bad reference is skipped now.

## Notes on the diff

Regenerated properly via `pnpm patch`/`pnpm patch-commit` rather than hand-editing the `.patch` file. The patch hash in `pnpm-lock.yaml` is a plain sha256 of the patch file, so I applied just that 3-line change instead of committing an unrelated full lockfile reformat; verified with `pnpm install --frozen-lockfile` ("lockfile is up to date, resolution step is skipped").

## Test plan

- `node --check` on the patched file.
- `pnpm install --frozen-lockfile` - clean, no drift.
- Confirmed against the real GitHub GraphQL API that a batch with one bad alias (`#8203`) still returns valid sibling data (`error.data`) for the rest.
- Reproduced both scenarios against a mocked Octokit: (1) a single-alias failure now still comments/labels the other real issues/PRs in the same batch, and (2) a genuine total failure (no partial data) is still caught by the outer wrapper instead of crashing.

## Not in this PR

Manually recreating the missed success-step comments/labels for the already-shipped `9.1.0-alpha.1` release - being handled separately. The Docusaurus docs sync for that release is done in ongov/design-system-developer-documentation-site#27.

### DIFF
--- a/patches/@semantic-release__github@12.0.9.patch
+++ b/patches/@semantic-release__github@12.0.9.patch
@@ -1,8 +1,17 @@
 diff --git a/lib/success.js b/lib/success.js
-index 48246f29a2162ce0abc0d044112de94bab39becb..e106ddf9ba078c9f75ef06ef1c358183654af40f 100644
+index 48246f29a2162ce0abc0d044112de94bab39becb..72edb4d289d19b482a6759ce04e2c02475e3afdd 100644
 --- a/lib/success.js
 +++ b/lib/success.js
-@@ -78,10 +78,20 @@ export default async function success(pluginConfig, context, { Octokit }) {
+@@ -14,7 +14,7 @@ import getReleaseLinks from "./get-release-links.js";
+ 
+ const debug = debugFactory("semantic-release:github");
+ 
+-export default async function success(pluginConfig, context, { Octokit }) {
++async function successImpl(pluginConfig, context, { Octokit }) {
+   const {
+     options: { repositoryUrl },
+     commits,
+@@ -78,13 +78,25 @@ export default async function success(pluginConfig, context, { Octokit }) {
      // Get associatedPRs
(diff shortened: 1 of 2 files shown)


muntalee (approved):
LGTM 🐐

erincarrieresontario (approved):
👍

PR #347, opened 2026-08-11
DS-2432: Add Next.js E2E and VRT coverage for `ontario-search-box` and `ontario-search-result-item`

muntalee:
## Summary
This PR adds focused Next.js test coverage for `ontario-search-box`, including new end-to-end checks and visual regression coverage for key interactive/autocomplete states across browsers.

## Changes
- Added a new E2E test suite for Next.js
- Added a new VRT test suite for Next.js
- Added VRT baseline snapshots (Chromium, Firefox, WebKit) for default, autocomplete, hover, focus, and active states
- Updated the Next.js search box component demo page used by tests
- Updated search box component source and docs to align with coverage changes


packages/ontario-design-system-component-library/src/components/ontario-search-box/readme.md
```diff
@@ -221,42 +261,11 @@ export class SearchAutocompleteComponent {
 </Tabs>
 ```
 
-### Slotted semantic and custom HTML suggestions
-
```
smorrisods, line 225: This removes the `"Slotted semantic and custom HTML suggestions"` and `"Grouping suggestions with static headers"` sections, the `"Slot content takes precedence over getSuggestions(query)..."` note here, the matching JSDoc on `getSuggestions`/`autocompleteSuggestionsUpdated` in `ontario-search-box.tsx` (and the generated `components.d.ts`/Angular `components.ts`), and the `"Example with custom slotted content"` section in `ontario-search-result-item/readme.md`.

Tracing through `ontario-search-box.tsx`'s `handleAutocompleteInput`, though, the slot-based path (`hasSuggestionSlotContent`, `decorateSlotSuggestionOptions`) is still fully implemented and still emits `autocompleteSuggestionsUpdated` — the code wasn't removed, just the docs describing it. That makes the new JSDoc (`"Emitted after asynchronous suggestions are updated."`) inaccurate, since the event also fires in slot mode.

Since this PR's stated purpose is adding test coverage, was removing slot-mode documentation intentional (e.g. deprecating it in favour of `getSuggestions` going forward), or is this fallout from focusing the new tests/demo page on the async path only? If it's still supported, let's restore the docs and JSDoc; if it's being deprecated, that feels like its own decision/PR rather than something to fold quietly into a testing PR.

packages/app-nextjs/src/app/components/ontario-search-box/page.tsx
```diff
@@ -1,70 +1,116 @@
...
-				<h3>Fr</h3>
-				<OntarioSearchBox language="fr" caption="Recherche en français"></OntarioSearchBox>
+				<div>
+					<h1>ontario-search-box</h1>
 
-				<h2>&apos;required&apos; Prop Variants</h2>
-				<h3>True</h3>
-				<OntarioSearchBox required={true} caption="Required field"></OntarioSearchBox>
+					{/*
+					Stencil reference (original)
```
smorrisods: The rewritten demo page drops the previous `language`, `required`, `hintText`, `performSearch`, and custom event handler prop-variant examples down to a single autocomplete demo, and leaves a large commented-out `"Stencil reference (original)"` JSX block in the file. Since this page doubles as the component's live reference/demo, losing those variant examples is a real reduction in demo coverage, and a commented-out block isn't something we'd want to ship — either restore the removed variants alongside the new autocomplete example, or clean out the dead comment if it's just leftover scaffolding.

packages/app-nextjs/tests/ontario-search-box/ontario-search-box.vrt.ts
```diff
@@ -0,0 +1,181 @@
...
+	const suggestions = search.locator('.ontario-search-autocomplete__suggestion-list');
+	const listHeight = await suggestions.evaluate((element) => element.scrollHeight);
+
+	await search.locator('.ontario-search__input-suggestion-container').evaluate((container, height) => {
+		container.style.paddingBottom = `${height}px`;
+	}, listHeight);
+};
+
+const gotoSearchPage = async (page: Page) => {
+	await page.goto('/components/ontario-search-box');
```
smorrisods: This duplicates the `beforeEach` setup in `ontario-search-box.e2e.ts` almost line-for-line (navigate, wait for `customElements.whenDefined`, wait for `componentOnReady`, poll for `performSearch`/`getSuggestions`). Since it's already extracted into a named function here, let's move it into the shared `../vrt-helpers` (or a small shared test-utils module) so both files import the same implementation instead of keeping two copies in sync.

smorrisods (changes requested):
Solid E2E/VRT coverage for the async autocomplete path, and nice use of the shared `expectVrtScreenshot`/`waitForInteractionPaint` helpers. Bigger issue: this PR also quietly removes documentation (and adjusts JSDoc) for the slot-based suggestion mode, even though that code path is still live and functional — that reads like scope creep rather than something that belongs in a "add test coverage" PR, so flagging it as the main thing to resolve before merge. Also flagged a shrunken demo page and some setup duplication between the e2e/VRT files.

---

packages/ontario-design-system-component-library/src/components/ontario-search-box/readme.md
muntalee: Traced `handleAutocompleteInput` and slot mode still fires `autocompleteSuggestionsUpdated`, so this was fallout, not a deprecation. Restored both readme sections, the slot-precedence note, and the JSDoc on `getSuggestions`/`autocompleteSuggestionsUpdated` in 28c15582. Also restored the slotted example in `ontario-search-result-item/readme.md`.

packages/app-nextjs/src/app/components/ontario-search-box/page.tsx
muntalee: Cleaned out the dead comment block in 35e90dcc.

packages/app-nextjs/tests/ontario-search-box/ontario-search-box.vrt.ts
muntalee: Extracted it into `gotoSearchBoxPage` in `tests/ontario-search-box/utils/goto-search-box-page.ts`, used by both files now in dc3f27a4.

erincarrieresontario (approved):
LGTM 👍

PR #349, opened 2026-08-11
DS-2666: Add Next.js E2E and VRT coverage for `ontario-in-page-navigation` and `ontario-in-page-navigation-item`

muntalee:
## Summary
This PR adds focused Next.js test coverage for Ontario In-Page Navigation, including new end-to-end checks and visual regression coverage for key default and interactive states across browsers.

This PR also includes dedicated Next.js coverage for Ontario In-Page Navigation Item, with separate E2E and VRT suites plus baseline snapshots for default, interaction, and clicked persistent states.

## Changes
- Added a new Next.js E2E test suite for Ontario In-Page Navigation
- Added a new Next.js VRT test suite for Ontario In-Page Navigation
- Added VRT baseline snapshots (Chromium, Firefox, WebKit) for default, hover, focus, and active states
- Added/updated the Next.js in-page-navigation demo page used by tests
- Updated the Next.js app index wiring to include navigation to the new component demo page
- Added a dedicated Next.js E2E test suite for Ontario In-Page Navigation Item in its own test folder
- Added a dedicated Next.js VRT test suite for Ontario In-Page Navigation Item in its own test folder
- Added Ontario In-Page Navigation Item VRT baseline snapshots (Chromium, Firefox, WebKit) for:
  - default item state
  - focused item link
  - hovered item link
  - clicked item link after reload with hash (persistent/visited flow)
- Refactored item-focused assertions out of the parent In-Page Navigation E2E suite so item behaviour is validated in the dedicated item suite


packages/app-nextjs/src/app/components/ontario-in-page-navigation/page.tsx
erincarrieresontario, line 1: Skip links point to a missing target. page.tsx renders <main> without id="skip-to-main", but both OntarioInPageNavigation instances use the component’s default skip target. Keyboard users will get a broken “Skip this page navigation” link.

packages/app-nextjs/tests/ontario-in-page-navigation-item/ontario-in-page-navigation-item.e2e.ts
erincarrieresontario, line 1: hmmm these E2E tests don’t cover the skip-link behavior that this fixture relies on.
These tests verify anchors and variants, but ontario-in-page-navigation.e2e.ts does not assert the skip link target exists or works

erincarrieresontario (approved):
A few comments, otherwise looks good :)

packages/app-nextjs/src/app/components/ontario-in-page-navigation/page.tsx
muntalee: Good catch — fixed in ab79a4d. Added `id="skip-to-main"` to the page's `<main>` so the skip link's default target (`skip-link-target` defaults to `skip-to-main`) resolves to a real element instead of a broken anchor.

packages/app-nextjs/tests/ontario-in-page-navigation-item/ontario-in-page-navigation-item.e2e.ts
muntalee: Added two new e2e tests in ab79a4d covering this: one asserting the skip link's href points at an element that exists on the page, and one that focuses the link via keyboard and presses Enter (the link is visually hidden until focus, so we drive it the same way a keyboard/AT user would) and asserts the URL hash updates. This also would have caught the missing `id="skip-to-main"` bug.

PR #350, opened 2026-08-11
DS-2691: Configure Style Dictionary primitive output under CSS layers

smorrisods:
## Summary
- Adds new Style Dictionary output platforms for the primitive token layer (`tokens/primitives/`): `primitives.css`, `primitives.scss`, `primitives.js`, and `primitives.ts`/`primitives.d.ts`.
- Wraps the CSS custom properties in `@layer ods-tokens` for deterministic cascade, and emits every primitive output under an `ods-` prefix (for example `--ods-space-400`, `$ods-radius-100`, `OdsBreakpointMedium`).
- Applies the `pxToRem`/`pxToEm` primitive value transforms (registered in DS-2690) per category: text-relative categories (space, font, line-height, letter-spacing, radius, border) convert px → rem; breakpoints convert px → em; everything else (elevation, motion, z-index) is left as authored.
- Adds `package.json` `exports` entries for the new primitive entry points (`./css/primitives`, `./scss/primitives`, `./primitives`) while leaving the existing `.`, `./css`, `./scss`, `./tokens` exports and their `dist/**/_variables.*`/`tokens.*` outputs byte-for-byte unchanged.

## Why
DS-2691 wires up the colour (DS-2688) and non-colour (DS-2690) primitive sources as consumable outputs. The new layered outputs ship additively so existing consumers of the flat token model are unaffected until they opt in to the new primitive entry points.

## Validation
- `pnpm test` (new `test/primitive-platforms.test.mjs` + existing suites): 18 tests passing, covering the `@layer` wrapping, `ods-` prefix, unit conversions, byte-for-byte legacy output, and `package.json` exports resolution.
- `node build.mjs`: legacy `dist/{css,scss,js,ts}` outputs remain byte-identical to the committed `legacy/` snapshot; new `primitives.*` files generated alongside them.
- `pnpm run lint:tokens`: 22 files checked, 0 warnings, 0 errors.

DS-2691

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/packages/ontario-design-system-design-tokens/scripts/config/primitive.config.ts
+++ b/packages/ontario-design-system-design-tokens/scripts/config/primitive.config.ts
@@ -11,12 +11,14 @@
  * everything else -> unitless/left as authored.
  */
 
+import type { Config } from 'style-dictionary/types';
+
 /** Shared value/name transforms applied ahead of each platform's naming transform. */
 const CSS_LIKE_TRANSFORMS = ['attribute/cti', 'name/kebab', 'color/hsl', 'size/pxToRem', 'size/pxToEm'];
 const JS_LIKE_TRANSFORMS = ['attribute/cti', 'name/pascal', 'color/hsl', 'size/pxToRem', 'size/pxToEm'];
 
 /** Style Dictionary config for the primitive layer's `primitives.*` outputs. */
-export const primitivePlatformsConfig = {
+export const primitivePlatformsConfig: Config = {
 	source: ['tokens/primitives/**/*.json'],
 	platforms: {
 		'css/primitives': {
--- a/packages/ontario-design-system-design-tokens/test/primitive-platforms.test.ts
+++ b/packages/ontario-design-system-design-tokens/test/primitive-platforms.test.ts
@@ -1,7 +1,7 @@
 /**
  * @file Tests for the primitive layer's Style Dictionary output platforms
  * (DS-2691): the `primitives.*` CSS/SCSS/JS/TS entry points built from
- * scripts/config/primitive.config.mjs, registered via the transforms in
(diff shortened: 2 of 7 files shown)


muntalee (approved):
LGTM @smorrisods 🐐

packages/ontario-design-system-design-tokens/test/primitive-platforms.test.ts
```diff
@@ -0,0 +1,69 @@
...
+import StyleDictionary from 'style-dictionary';
+
+import { primitiveTransforms } from '../scripts/lib/transforms.ts';
+import { primitivePlatformsConfig } from '../scripts/config/primitive.config.ts';
+
+const packageRoot = path.dirname(path.dirname(fileURLToPath(import.meta.url)));
+
+// Style Dictionary resolves its `source` glob relative to process.cwd().
+beforeAll(() => {
+	process.chdir(packageRoot);
```
erincarrieresontario, line 20: process.chdir() is not restored after the test, this changes global process state and never changes it back. That can leak into other Vitest files running in the same worker.

packages/ontario-design-system-design-tokens/scripts/config/primitive.config.ts
erincarrieresontario, line 1: Could we add a quick README note for the new primitive exports? Since ./css/primitives, ./scss/primitives, and ./primitives are now public entry points, it’d be helpful to show when to use them and mention that the CSS version is wrapped in @layer ods-tokens.

erincarrieresontario (approved):
Just a few comments, otherwise looks good 👍

PR #353, opened 2026-08-11
DS-2692: Add Figma and downstream/Tailwind JSON exports for primitives

smorrisods:
## Summary

Adds Figma, downstream, and Tailwind JSON exports for the primitive token layer (DS-2692), completing the DS-2685 primitive-layer Epic's export tooling alongside the SD platform work from DS-2691 (#350):

- All three exports — a Figma Variables-compatible bundle, a typed downstream JSON tree, and a Tailwind `theme.extend` mapping — are Style Dictionary custom formats/platforms built from a single Style Dictionary instance in `build.mjs`, alongside the primitive CSS/SCSS/JS/TS platforms. This matches the GC Design System's own token export tooling ([github.com/cds-snc/gcds-tokens](https://github.com/cds-snc/gcds-tokens)), which produces every platform — including Figma — from one `style-dictionary build` config, rather than a separate bespoke script per output.
- Every export reads a token's export type (`borderWidth`, `spacing`, `fontFamilies`, `boxShadow`, etc.) directly off the resolved Style Dictionary token. The `type` an author declares on a primitive's source JSON (`tokens/primitives/`) is the single source of truth — nothing infers a type from a token's path or value.
- `scripts/lib/token-types.ts` defines the closed `TOKEN_TYPES` vocabulary (a `TokenType` union plus an `isTokenType` guard) that every declared `type` must belong to, and the header comment in `scripts/lib/token-export.ts` documents how to add a new type to it.
- `scripts/lib/token-tooling.ts`'s linter enforces that vocabulary: a `missing_type` error for any leaf token with a `value` but no declared `type`, and an `invalid_type` error for a declared `type` outside `TOKEN_TYPES`. This lint runs once, up front, in `build.mjs`, before any platform builds, so a missing or invalid type hard-fails the build rather than silently shipping an untyped or mistyped export.
- `scripts/lib/token-export.ts` reconstructs a nested token tree from a Style Dictionary dictionary's flat `allTokens` list — typed `{ value, type }` nodes for the Figma/downstream exports, plain values for Tailwind — shared by every format registered in `scripts/lib/export-platforms.ts`, so Figma and downstream are built from the exact same resolved tokens.
- Added `test/export-platforms.test.ts` (Figma/downstream/Tailwind exports: Core-only scoping, authored-type spot-checks, Figma/downstream parity, Tailwind colour-scale flattening) and extended `test/tokens.test.ts` with `missing_type`/`invalid_type` regression tests and vocabulary coverage for `token-types.ts`.

## Why

DS-2692 closes out the Epic's export tooling: designers need a Figma Variables-compatible bundle, and downstream/Tailwind consumers need typed JSON built from the same primitives, all staying in sync automatically via the standard build. Driving every export's typing off one authored, Style-Dictionary-resolved field against a closed vocabulary — rather than reconstructing it per export from a token's path/value, or accepting any string — means there is exactly one place a token's export type is decided, and a build-time guarantee that it's always present and valid.

## Validation

- `pnpm run build` — lints, then builds the primitive `primitives.*` outputs and the new `exports/figma`, `exports/downstream`, `exports/tailwind` JSON, with no errors and no untyped tokens.
- `pnpm test` — 21/21 tests pass.
- `pnpm run lint:tokens` — 22 files checked, 0 warnings, 0 errors.
- Manually confirmed `build.mjs` exits non-zero when a primitive token is missing its `type`, or when a declared `type` isn't in `TOKEN_TYPES`.

### DIFF
--- a/packages/ontario-design-system-design-tokens/build.mjs
+++ b/packages/ontario-design-system-design-tokens/build.mjs
@@ -10,6 +10,7 @@ import StyleDictionary from 'style-dictionary';
 
 import { primitiveTransforms } from './scripts/lib/transforms.ts';
 import { primitivePlatformsConfig } from './scripts/config/primitive.config.ts';
+import { exportPlatformsConfig } from './scripts/lib/export-platforms.ts';
 
 // Register the primitive value transforms so they are available to the
 // primitive output platforms configured below (DS-2691 / PR 5).
@@ -54,3 +55,11 @@ const primitivesDictionary = new StyleDictionary(primitivePlatformsConfig);
 await primitivesDictionary.hasInitialized;
 await primitivesDictionary.buildAllPlatforms();
 console.log('Built primitive layer output (primitives.*).');
+
+// Build the downstream/Tailwind JSON exports (DS-2692). Additive, alongside
+// the primitive layer outputs above; consumed by Figma/downstream/Tailwind
+// tooling rather than published as package entry points.
+const exportsDictionary = new StyleDictionary(exportPlatformsConfig);
+await exportsDictionary.hasInitialized;
+await exportsDictionary.buildAllPlatforms();
+console.log('Built downstream/Tailwind JSON exports (exports/downstream, exports/tailwind).');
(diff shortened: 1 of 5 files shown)


muntalee (changes requested):
Small note: the description says these tokens all have an explicit type field, but I don't see one in any of the new JSON files, looks like they're relying on the auto-detection in #353 instead. Not a blocker, just want to flag it in case the description should be updated.

PR #354, opened 2026-08-11
DS-2603: Add new process for shipping css for react

erincarrieresontario:
- Right now, if a team wants to use our React components in Next.js, they also need our CSS, but we only ship Sass source files, not actual CSS. That means every team has to have Sass compiling properly in their own build, and for a lot of Next.js setups (especially with Turbopack), that just doesn't work cleanly. Some teams get stuck entirely.
- This MR fixes that by precompiling our Sass into a plain `theme.css` file and shipping it with the React package. Now teams can just do:
`import '@ongov/ontario-design-system-component-library-react/theme.css';`

**How it works**
- Added a small script that runs after our normal build and compiles the theme's Sass into real CSS.
- Along the way, I found and fixed a bug where our Sass has a variable that controls where font/image URLs point [$asset-base-path] and its default only makes sense inside our own repo. Once compiled and shipped elsewhere, those paths would've broken and fonts would 404. This was fixed by baking in the [/assets] path we already tell consumers to copy their fonts into.
- Added the new `theme.css` export to the package so it's actually importable. The old Sass path is still there too, nothing's removed, this is additive.

### DIFF
--- a/packages/ontario-design-system-component-library-react/scripts/build-css.mjs
+++ b/packages/ontario-design-system-component-library-react/scripts/build-css.mjs
@@ -1,0 +1,67 @@
+import { compile, compileString } from 'sass-embedded';
+import { NodePackageImporter } from 'sass-embedded';
+import { fileURLToPath } from 'node:url';
+import path from 'node:path';
+import { mkdir, writeFile } from 'node:fs/promises';
+
+/**
+ * Compiles the Ontario Design System theme Sass entry point into a single,
+ * distributable CSS file.
+ *
+ * This is a monolithic output (one `theme.css` containing base styles, global
+ * styles, and all component styles) chosen as the initial CSS output strategy
+ * so that Next.js (and other) consumers can `import` a ready-to-use stylesheet
+ * without needing a Sass compiler configured in their own build pipeline.
+ *
+ * The `NodePackageImporter` is required here because the copied `theme.scss`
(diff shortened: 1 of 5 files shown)


packages/ontario-design-system-component-library-react/scripts/build-css.mjs
```diff
@@ -0,0 +1,67 @@
...
+		importers: [new NodePackageImporter()],
+		url: new URL(`file://${outDir}/__theme-entry.scss`),
+	});
+
+	await mkdir(outDir, { recursive: true });
+	await writeFile(outFile, result.css, 'utf8');
+
+	if (result.sourceMap) {
+		await writeFile(`${outFile}.map`, JSON.stringify(result.sourceMap), 'utf8');
+	}
```
muntalee, line 62: `theme.css.map` is generated but nothing links it to `theme.css` — sass-embedded doesn't add the `sourceMappingURL` comment automatically (per its own docs), so devtools won't pick up the map as-is. Might want to append `\n/*# sourceMappingURL=theme.css.map */` to `result.css` before writing it out.

muntalee (changes requested):

packages/ontario-design-system-component-library-react/scripts/build-css.mjs
erincarrieresontario: Good catch @muntalee , thanks :) Updated!

PR #355, opened 2026-08-11
DS-2693: Add CI guardrails for primitive token builds

smorrisods:
## Summary

Adds three CI guardrails around the primitive token build, so future changes to `tokens/primitives/**` are caught automatically instead of relying on manual review:

1. **Alias/integrity linter in CI** — the existing `lint-tokens.mjs` script (from DS-2687, already run as a pre-flight inside the Figma export) is now wired in as its own step in `build.yml`, so a `missing_alias_target` error fails the build independently of any export.
2. **Token output tests** — a new `test/token-output-snapshot.test.mjs` builds the primitive `css/primitives` platform and asserts the resulting token count and names, per category, against a committed known-good snapshot (`test/fixtures/primitive-token-snapshot.json`; 318 tokens across 11 categories). This catches an accidental deletion/rename of a primitive token, which still produces valid (just smaller/different) Style Dictionary output and wouldn't otherwise fail a build.
3. **CSS size budget** — a new step in `build.yml` checks `dist/css/primitives.css`'s byte size against an agreed budget (25 KiB, current size ~15.5 KB) and fails the build if exceeded, to catch unintentional bloat.

A `pnpm run generate:token-snapshot` script is included to regenerate the fixture after an intentional token change; the PR description asks reviewers to confirm the regenerated diff matches the intended change before merging.

## Why

DS-2693 closes out the remaining "Develop" work for the DS-2685 Epic by adding automated guardrails around the primitive token pipeline built in DS-2690/DS-2691/DS-2692, so drift/regressions in the primitive layer are caught in CI rather than downstream (e.g. in a Figma sync or a consuming app).

Out of scope (per the ticket): visual regression testing (separate ticket) and Semantic/Component token guardrails (later Epics).

## Validation

- `pnpm test` — all suites pass, including the 13 new token-output-snapshot assertions.
- `pnpm run lint:tokens` — 0 errors, 0 warnings.
- Manually verified each guardrail catches real breakage (per the ticket's Test Notes), then reverted:
  - Deleted a token (`border.width.300`) → `token-output-snapshot.test.mjs` failed with a clear diff; restored.
  - Introduced a broken alias reference → `lint-tokens.mjs` failed with `missing_alias_target`; removed.
- Confirmed `pnpm run generate:token-snapshot` reproduces the committed fixture byte-for-byte against the current token set.

## Stack

Part of the DS-2685 Epic stack: #351 ← #299 ← #350 ← #353 ← #355 (this PR).


muntalee (approved):
LGTM @smorrisods 🐐

PR #357, opened 2026-08-14
DS-2584: add playwright and VRT automation DS-2586 for ontario-callout

tresajm:
The following are checked under these tests.

### Verify DOM structure, classes, and prop-driven behaviour:
- Default render (renders) — confirms hydration + default classes (ontario-callout, ontario-border-highlight--teal).
- Dynamic [headingType
- Dynamic [highlightColour]
- Dynamic [headingContentType]

### Accessibility Tests
- Scans a callout with static [heading-content]/[heading-content-type] attributes.

### Visual Regression Tests
- Default render screenshot.
- Screenshot after each [headingType]

Currently a bug is entered for accessibility violation
https://ontariodotca.atlassian.net/browse/DS-2774


smorrisods:
@tresajm, please review the title of this PR and make sure it follows our naming guidelines.

packages/ontario-design-system-component-library/src/components/ontario-callout/test/ontario-callout.e2e.ts
```diff
@@ -1,65 +1,109 @@
...
+	});
 
-// 		beforeEach(async () => {
-// 			page = await newE2EPage();
-// 			await page.setContent('<ontario-callout></ontario-callout>');
-// 			component = await page.find('ontario-callout');
-// 			element = await page.find('ontario-callout >>> div');
-// 		});
+	test.skip('has no detectable accessibility violations', async ({ page }) => {
+		await page.setContent(`
```
smorrisods, line 19: This a11y test is skipped with no explanation (no `// TODO`, no linked issue). Tracing through, the last test in this same `describe` block (`'has no accessibility violations after property updates'`) covers overlapping ground — it sets `headingContentType`/`headingContent` and runs an axe scan — and isn't skipped. If this skipped test is now redundant, let's drop it rather than leave a silently-skipped test sitting in the suite; if it's actually blocked on something real, a comment linking the reason/issue would help whoever looks at this next.

packages/ontario-design-system-component-library/src/components/ontario-callout/test/ontario-callout.vrt.ts
```diff
@@ -0,0 +1,83 @@
...
+
+			const heading = component.locator('h2');
+
+			await expect(heading).toHaveText('Get notified');
+			await expect(heading).toHaveClass(/ontario-callout__title/);
+
+			await component.evaluate((el: HTMLOntarioCalloutElement) => {
+				el.headingContentType = 'html';
+				el.headingContent = '#';
+				el.headingContent = '<a href="#">Get notified</a>';
```
smorrisods, line 76:
```ts
el.headingContentType = 'html';
el.headingContent = '#';
el.headingContent = '<a href="#">Get notified</a>';
```
The `el.headingContent = '#';` line gets immediately overwritten by the next line and never takes effect — looks like a leftover from editing/debugging. Let's remove it so the test only sets the value it actually asserts against.

smorrisods (changes requested):
Good to see `ontario-callout` get real Playwright coverage after being fully commented out. A couple of small things worth cleaning up before merge: a leftover, immediately-overwritten line in the VRT file, and an unexplained skipped a11y test that looks redundant with a later one. Details inline.

---
🤖 Beep boop — Ghostwritten by CoPilot — Human judgment (and sarcasm) still pending.

packages/ontario-design-system-component-library/src/components/ontario-callout/test/ontario-callout.vrt.ts
```diff
@@ -0,0 +1,83 @@
...
+		await page.setContent('<ontario-callout></ontario-callout>');
+
+		const component = page.locator('ontario-callout');
+		const element = component.locator('div').first();
+
+		await expect(component).toBeVisible();
+		await expect(component).toHaveClass(/hydrated/);
+		await expect(element).toHaveClass(/ontario-callout/);
+		await expect(element).toHaveClass(/ontario-border-highlight--teal/);
+		await expect(page).toHaveScreenshot();
```
muntalee: This and the other `toHaveScreenshot()` calls in this file screenshot the whole `page` with no `animations: 'disabled'` / `caret: 'hide'`. Every other `*.vrt.ts` in the library scopes to the component host and disables animations/caret for determinism (see `ontario-back-button.vrt.ts`). Let's match that pattern here.

packages/ontario-design-system-component-library/src/components/ontario-callout/test/ontario-callout.vrt.ts
```diff
@@ -0,0 +1,83 @@
...
+
+			const goldElement = component.locator('.ontario-callout');
+			await expect(goldElement).toHaveClass(/ontario-border-highlight--gold/);
+
+			await component.evaluate((el: HTMLOntarioCalloutElement) => {
+				el.highlightColour = 'lime';
+			});
+
+			await expect(goldElement).toHaveClass(/ontario-border-highlight--lime/);
+			await expect(page).toHaveScreenshot();
```
muntalee: This only screenshots the final `lime` state. The `gold` state above is asserted at the DOM level but never actually captured, unlike the `headingType` test which screenshots after each change. Worth adding a screenshot after the `gold` change too.

muntalee (changes requested):
Good to see real Playwright coverage here too, on top of Scott's comments a couple more things worth fixing on the VRT side.

PR #363, opened 2026-08-17
DS-2582: added vrt tests for ontario-aside

tresajm:
This test is written to validate the visual regressions

- CSS/style validation tests — assert computed styles directly (border widths/style, colour changes, default colour fallback) rather than just visual output.
- Visual regression tests (VRT) —  to catch unintended visual changes across colour variants and states, compared against baseline snapshots.


smorrisods:
The title of the PR should follow the format:

> JiraIssue-ID: Descriptive human readable title

packages/ontario-design-system-component-library/src/components/ontario-aside/test/ontario-aside.vrt.ts
```diff
@@ -0,0 +1,167 @@
...
+		await page.waitForChanges();
+
+		host = page.locator('ontario-aside');
+		expect(page).toHaveScreenshot();
+	});
+
+	test('applies and updates border-highlight class when highlight-colour changes', async ({ page }) => {
+		// initial class on inner <aside>
+		const hasPurple = await host.evaluate(
+			(el: Element) => !!el.shadowRoot?.querySelector('aside')?.classList.contains('ontario-border-highlight--purple'),
```
muntalee: Missing `await` here. The screenshot can still be capturing when the test body mutates the DOM right after, e.g. in "applies and updates border-highlight class..." the before/after PNGs are byte-identical and both show the `lime` state, not the `purple` baseline. Also use `expectVrtScreenshot()` from `vrt-helpers.ts` instead of bare `toHaveScreenshot()`, same as the rest of the vrt suite (see #364).

muntalee (changes requested):
Looks good overall, nice coverage on the highlight-colour and content variants. Just needs one fix before merging, left a comment on the line. Also needs a rebase 🐐

PR #366, opened 2026-08-17
DS-1813: Autogenerated web component error messages

muntalee:
## Summary

This PR provides a proof of concept for generating consistent web component error messages from runtime-defined allowed values.

## Changes

- Added a shared utility for formatting allowed property values in validation warnings.
- Added shared invalid-prop warning handling for consistent error messages.
- Updated several existing components to use the shared warning pattern.
- Applied the pattern across the generated icon components, using their shared runtime colour and size values.
- Added focused unit tests for the warning formatter.

## Scope

This PR is intentionally scoped as a proof of concept for DS-1813. The approach is currently demonstrated through several components, with the most complete adoption being in the icon components and their generator.
\
The ticket describes extracting values from a TypeScript type. Because TypeScript types are not available at runtime, this implementation uses `const` value arrays as the runtime source of truth while preserving the corresponding TypeScript types. These values are then reused for validation and error-message generation, reducing duplicated and potentially outdated warning text.
\
The approach should be reviewed before expanding it across the remaining component library.

### DIFF
--- a/packages/ontario-design-system-component-library/src/utils/validation/invalid-prop-warning.spec.ts
+++ b/packages/ontario-design-system-component-library/src/utils/validation/invalid-prop-warning.spec.ts
@@ -1,0 +1,58 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import { formatAllowedValues } from '../helper/utils';
+import { printInvalidPropWarning } from './invalid-prop-warning';
+
+describe('formatAllowedValues', () => {
+	it('joins values with and by default', () => {
+		expect(formatAllowedValues(['vertical', 'horizontal'])).toEqual('vertical and horizontal');
+	});
+
+	it('joins values with or when requested', () => {
+		expect(formatAllowedValues(['string', 'html'], 'or')).toEqual('string or html');
+	});
+
+	it('formats numeric allowed values', () => {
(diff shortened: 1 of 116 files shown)


packages/ontario-design-system-component-library/src/utils/validation/invalid-prop-warning.spec.ts
erincarrieresontario, line 1: The helper tests only check substrings, not the full warning output. Since this branch is standardizing message text, it might be worth snapshotting or asserting the full console.warn output for at least one “invalid type” and one “invalid value” case.

erincarrieresontario (commented):
This branch migrates button/card/callout/aside/icons, but there are still similar warning patterns in components like ontario-summary-list.tsx, ontario-in-page-navigation.tsx, and ontario-loading-indicator.tsx. Could be follow-up work, but worth noting if DS-1813 is intended to cover all shared invalid prop warnings.

packages/ontario-design-system-component-library/src/utils/validation/invalid-prop-warning.spec.ts
muntalee: Fixed in f7e33f1e. For now this covers the two main cases (invalid type, invalid value) — can extend to more later if needed.

PR #369, opened 2026-08-20
DS-2725: made accordion text content searchable

erincarrieresontario:
**The problem**
Before this change, closed accordion sections used display: none to hide their content. That's a totally normal way to hide things, but it has one downside: the browser's Find-in-page search completely ignores anything hidden with display: none. So if a user pressed Ctrl+F looking for a word that only existed inside a closed accordion section, the browser would never find it — even though the answer was right there on the page.

**The fix**
We swapped display: none for a newer HTML attribute called hidden="until-found". It works almost identically from a visual standpoint (content still looks hidden), but the browser treats it differently:

- The content stays in the page's search index even while visually collapsed.
- If the browser finds a search match inside a closed section, it automatically expands that section and fires an event called beforematch to let us know it did that.
- We listen for that beforematch event and update our own component state (the open/closed tracking, the chevron icon, aria-expanded, etc.) so everything stays in sync — the component doesn't get confused about whether a section is open or closed.

For older browsers that don't support hidden="until-found" yet, we detect that automatically and fall back to the old display: none-style hiding, so nothing breaks for anyone.


muntalee (approved):
LGTM 🐐

andrewpeart (approved):
👍🏻

PR #375, opened 2026-08-24
DS-1913: Fix mixed-language footer links in the Angular POC

muntalee:
## Summary

The ticket reported the Angular POC translating only some footer links when switching to French. Ontario components read `<html lang>` as they hydrate, and `index.html` hard-coded `lang="en"` — so on a French page the copyright line hydrated as English while the app's own links rendered in French. This PR resolves `<html lang>` from the URL before Angular bootstraps and keeps it aligned on toggle.

## Changes

- Set `<html lang>` from the URL in `index.html`, before Angular bootstraps
- Route all language switching through one `setLanguage()`, replacing the
  `onLangChange` subscription and the toggle's own logic
- Derive routes and footer links from `currentLang` instead of re-reading the URL
- Move footer links from an inline template literal into a `footerLinks` getter
- Bind `language` on `ontario-header` and `ontario-footer`
- Drop now-unused `Renderer2`, `AppConfigService` and `get-language` utils
- Add a footer regression test covering built-in links following `language`

## Scope

`app-angular` only, plus one component-library test. No component source changed.

## Verification

https://github.com/user-attachments/assets/7362dfca-43e1-4d98-94ba-54baef5686e6

### DIFF
--- a/packages/app-angular/src/index.html
+++ b/packages/app-angular/src/index.html
@@ -11,6 +11,11 @@
 		<link rel="icon" href="./assets/favicons/favicon-48x48.png" type="image/png" />
 		<link rel="icon" href="./assets/favicons/apple-touch-icon.png" type="image/png" />
 		<link rel="manifest" href="./assets/manifest.json" />
+		<script>
+			// Ontario components read <html lang> as they hydrate, which happens before Angular can
+			// set it, so the language is resolved from the URL here first.
+			document.documentElement.lang = window.location.hash.includes('/fr/') ? 'fr' : 'en';
+		</script>
 	</head>
 	<body>
 		<app-root></app-root>
--- a/packages/ontario-design-system-component-library/src/components/ontario-footer/test/ontario-footer.spec.tsx
+++ b/packages/ontario-design-system-component-library/src/components/ontario-footer/test/ontario-footer.spec.tsx
@@ -57,6 +57,20 @@ describe('ontario-footer', () => {
 		expect(inlineLinks?.[2].textContent?.trim()).toBe('Contact');
 	});
 
+	it('updates default links when language changes', async () => {
+		const page = await render('<ontario-footer type="default" language="en"></ontario-footer>');
+		const footer = page.root as HTMLElement & { language: 'en' | 'fr' };
+
+		footer.language = 'fr';
+		await page.waitForChanges();
+
+		const inlineLinks = page.root?.shadowRoot?.querySelectorAll(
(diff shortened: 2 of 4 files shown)


packages/app-angular/src/index.html
```diff
@@ -11,6 +11,11 @@
 		<link rel="icon" href="./assets/favicons/favicon-48x48.png" type="image/png" />
 		<link rel="icon" href="./assets/favicons/apple-touch-icon.png" type="image/png" />
 		<link rel="manifest" href="./assets/manifest.json" />
+		<script>
+			// Ontario components read <html lang> as they hydrate, which happens before Angular can
+			// set it, so the language is resolved from the URL here first.
+			document.documentElement.lang = window.location.hash.includes('/fr/') ? 'fr' : 'en';
```
erincarrieresontario: That works for #/fr/demarrer, but the Angular POC URL helper can generate #fr/demarrer without the slash after #. In that case, the page would set <html lang="en"> even on a French route. Could we normalize the hash or check for both #/fr/ and #fr/

packages/ontario-design-system-component-library/src/components/ontario-footer/test/ontario-footer.spec.tsx
```diff
@@ -57,6 +57,20 @@ describe('ontario-footer', () => {
 		expect(inlineLinks?.[2].textContent?.trim()).toBe('Contact');
 	});
 
+	it('updates default links when language changes', async () => {
```
erincarrieresontario, line 60: this verifies changing footer.language from en to fr, which is good, but the bug is specifically pre-hydration language resolution in the Angular app. An Angular/app-level test for loading a French hash URL would be more valuable.

erincarrieresontario (commented):
Looks good @muntalee , just a few comments :)

packages/app-angular/src/index.html
muntalee: Fixed in 2ba31c21

packages/ontario-design-system-component-library/src/components/ontario-footer/test/ontario-footer.spec.tsx
muntalee: Fair point, this only tests the component side. app-angular doesn't have any app-level/e2e test setup yet to cover the actual pre-hydration behaviour — mind if we track that as a follow-up ticket instead of bundling new test infra into this fix?

PR #386, opened 2026-08-31
DS-2013: Fix Complete Styles package not working in Angular application

muntalee:
## Summary
The Complete Styles package couldn't be used in the Angular application. The ticket called out an issue with file-watching in the build process, and that's real, but tracking it down also surfaced a few build/Sass bugs that were the actual reason Angular couldn't compile the package at all. All of it needed fixing to actually resolve the Angular issue, so it's included here together.

## Changes
- **Why Angular couldn't compile it at all**
  - Complete Styles' package exports had two entries that could both match the same file, which Angular's Sass compiler couldn't resolve ("ambiguous resolution" error).
  - The package's own build wasn't catching Sass errors, so it kept reporting success even when the compiled output was broken or stale — this hid the export bug above, plus a few Sass naming clashes between components, from ever being caught.
  - Fixed a handful of components (mostly the header family, plus back-button/step-indicator) with accidentally clashing Sass variable and mixin names, and removed some redundant re-exports.
- **File-watching, as described in the ticket**
  - The package's `scss` script pointed at a folder structure that hasn't existed since it was reorganized, so it failed immediately.
  - Its `watch` task only watched its own `src/`, which the build deletes as its last step, so it never had anything to watch.
  - Rewired `watch` to follow the real upstream sources (global-styles and component-library) and rebuild automatically when they change.
- Updated the global-styles README to match the corrected exports.

## Testing
- Rebuilt Complete Styles and confirmed the ambiguous export and build swallowing issues are gone
- Verified `pnpm run watch` does a real build and rebuilds automatically when an upstream source file changes
- Manually wired Complete Styles into an Angular POC and confirmed it now compiles and renders correctly, resolving the reported issue
- Also spot-checked the same setup in React while testing — not part of this ticket, but confirming it still renders correctly too

### DIFF
--- a/packages/ontario-design-system-complete-styles/gulpfile.js
+++ b/packages/ontario-design-system-complete-styles/gulpfile.js
@@ -31,11 +31,21 @@ const compileSass = (input, outputFile, options) => {
 		sassOptions.sourceComments = true;
 	}
 
-	return src(input, { sourcemaps: options.sourcemaps })
-		.pipe(sass(sassOptions).on('error', sass.logError))
+	const sassStream = sass(sassOptions);
+	const outputStream = src(input, { sourcemaps: options.sourcemaps })
+		.pipe(sassStream)
 		.pipe(gulpif(options.compress, concat(`${outputFile}.min.css`), concat(`${outputFile}.css`)))
 		.pipe(gulpif(options.compress, minify()))
 		.pipe(dest(paths.output.theme, { sourcemaps: options.sourcemaps ? '.' : false }));
+
+	// gulp-sass logs compile errors but doesn't fail the pipeline by default; forward the
+	// error onto the returned stream so gulp treats a broken Sass compile as a build failure.
+	sassStream.on('error', (error) => {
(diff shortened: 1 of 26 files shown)


packages/ontario-design-system-complete-styles/gulpfile.js
erincarrieresontario, line 1: Not sure this is safe as-is: the Sass build tasks in this file call processSass(...) but never return the stream(s). In Gulp, that means the task can finish before the CSS is actually written, which is exactly the kind of thing that makes deploys flaky. I’d return the stream/parallel result from sass:build and sass:minify so the build waits for the pipeline to complete.

muntalee: Just double checked it and this is actually already handled. `compileSass()` returns `outputStream` (with the `sassStream.on('error', ...)` forwarding onto it), and `sass:build`/`sass:minify` both `return compileSass(...)` directly, so Gulp waits on the pipeline and a Sass error will fail the task. Let me know if you still see a gap!

PR #356, opened 2026-08-12
Add VS Code dev container configuration

smorrisods:
## Summary

Adds a VS Code Dev Container (`.devcontainer/`) so contributors get one complete, native environment for building, linting, unit testing, and running e2e/vrt tests — no separate Docker Compose scripts required, and no Docker-in-Docker layering.

## What's included

- **Base image:** `mcr.microsoft.com/devcontainers/base:ubuntu-24.04` + the official `node` dev container feature — gives a pre-configured non-root `vscode` user (with passwordless `sudo`), git, and VS Code Server compatibility out of the box.
- **`postCreate.sh`:** dynamically resolves the exact Node.js (`.nvmrc`) and pnpm (`package.json`'s `packageManager`) versions to install at container-creation time, so there's a single source of truth instead of duplicating version numbers in `devcontainer.json`.
- **Playwright browsers installed natively** (matching the exact `@playwright/test` version pinned in the component library's `package.json`), so e2e/vrt tests run directly in the container.
- **Named volumes** shadow the root `node_modules`, pnpm's store (pnpm 11's default workspace-relative `.pnpm-store` location), and the Playwright browser cache — so Linux-built native binaries never overwrite a Mac/Windows host's own copies, and persist across container rebuilds.
- Forwards common local dev server ports (`3000` app-nextjs, `4200` app-angular, `5173` app-react) and adds recommended extensions (ESLint, Prettier, Stencil, Angular, EditorConfig, Code Spell Checker, Playwright).

## Motivation

Gives Mac/Windows/Linux contributors a consistent, ready-to-use dev environment without manually installing Node/pnpm/Playwright browsers on the host, and avoids the existing Docker Compose Playwright runner's need for host Docker/extra script layers for day-to-day development.

## Testing

Verified end-to-end using the `@devcontainers/cli` (`build` → `up` → `exec`):
- `pnpm install` completes successfully as the non-root `vscode` user.
- `pnpm run build-libs` succeeds across all 7 projects.
- Full Playwright e2e suite passes natively in-container (139 passed).
- Unit tests pass (285 passed, 2 skipped).
- Confirmed the named-volume shadowing works as intended: real dependency/cache content lives only in the volumes, while the host-side bind-mounted repo directory only shows empty placeholder folders.

PR #388, opened 2026-09-03
DS-2836: Bind app-angular and app-react dev servers to all interfaces

muntalee:
## Summary

`ng serve` and `vite` both default to binding on `localhost`, which resolves to the IPv6 loopback address (`::1`) only inside most Linux container network stacks. Port-forwarding tools (Dev Containers, Codespaces, the existing Docker Compose Playwright runner) connect over IPv4, so the connection is refused and the page hangs on loading indefinitely. `next dev` is unaffected since it already binds to a wildcard address covering both IPv4 and IPv6.

Found while testing the [VS Code Dev Container PR (#356)](https://github.com/ongov/ontario-design-system/pull/356), but not specific to it — this affects any containerized workflow for this repo (Codespaces, the Docker Compose Playwright runner, CI).

## Changes

- `app-react/vite.config.ts`: add `server.host: true` so Vite binds to all interfaces, matching Next.js's existing default behaviour
- `app-angular/package.json`: add `--host 0.0.0.0` to the `serve`/`start` scripts
- `.gitignore`: add missing `.pnpm-store/` entry

## Scope

`app-angular` and `app-react` only. No production build/deployment behaviour is affected — `ng build`/`vite build` outputs are unchanged.


packages/app-angular/package.json
```diff
@@ -4,8 +4,8 @@
 	"type": "module",
 	"scripts": {
 		"ng": "ng",
-		"start": "pnpm run clean:full && pnpm i && ng serve",
-		"serve": "ng serve",
+		"start": "pnpm run clean:full && pnpm i && ng serve --host 0.0.0.0",
```
smorrisods: Let's have this reference `pnpm run serve` so there is only one location for this change.

packages/app-react/vite.config.ts
```diff
@@ -8,6 +8,14 @@ const workspaceRoot = path.resolve(__dirname, '../..');
 export default defineConfig({
 	base: './',
 	plugins: [react()],
+	server: {
+		// Bind to all interfaces (not just IPv6 localhost) so the dev server is
+		// reachable from Docker/devcontainer/Codespaces port forwarding, which
+		// connects over IPv4. Without this, `vite`'s default `localhost` binding
+		// resolves to `::1` only in most container network stacks, and the
+		// forwarded port hangs indefinitely.
+		host: true,
```
smorrisods, line 19: With this is binding to `0.0.0.0` needed?

smorrisods (changes requested):

packages/app-angular/package.json
muntalee: Updated start to call `pnpm run serve` instead of duplicating the `--host 0.0.0.0` flag in 2758103b.

packages/app-react/vite.config.ts
muntalee: Yup, it's still needed. `host: true` is just Vite's shorthand for 0.0.0.0. Both `ng serve` and Vite default to `localhost`, which only listens on IPv6 in containers — so both needed this fix, just done differently per tool.

smorrisods (approved):
LGTM 🚀

erincarrieresontario (approved):
Looks good Munta!

PR #364, opened 2026-08-17
DS-2711: added e2e and vrt tests for Blockquote component

tresajm:
### **Test Summary**

This test suite validates the rendering, content handling, styling behavior, and accessibility of the  Blockquote component and VRT tests

### **Coverage Areas**

**Component Rendering & Hydration**

Verifies that the component renders successfully and hydrates correctly in the browser environment.

**Blockquote  Content Rendering**

Confirms that the supplied Blockquote  text is displayed correctly.
Validates fallback behaviour by ensuring the component uses the host element's textContent when a Blockquote prop is not provided.

**Conditional Styling**

Tests application of the short Blockquote  CSS class when the Blockquote  length is 140 characters or less.
Ensures the class is not applied when the Blockquote  exceeds 140 characters.

**Optional Metadata Rendering**

Verifies that attribution and byline elements are not rendered when corresponding data is absent.
Confirms correct rendering of:

- Attribution only
- Byline only
- Both attribution and byline together

**Accessibility Validation**

Runs automated accessibility checks using Axe to ensure the component has no detected accessibility violations.
Overall Validation

**Visual Regresssion Tests**

The tests provide comprehensive coverage of the Blockquote  component's core functionality, including rendering, dynamic content handling, conditional presentation logic, fallback behavior, optional metadata display, compliance with accessibility standards and Visual Regression tests.

PR #396, opened 2026-09-09
DS-2766: Add Playwright E2E & VRT tests for ontario-button

tresajm:
**Summary:**
The `ontario-button.e2e.ts` test suite verifies the complete user-facing behaviour of the Ontario button component. It confirms that the component renders with the correct default button type, styling, label, native HTML attributes, and accessibility information. It also verifies that button properties can be updated dynamically, including the label, accessible name, element ID, visual button type, and native HTML button type.

The suite tests both action-button and link behaviour, ensuring that an `href` renders the component as an anchor with the correct link attributes. It checks all supported `button styles and HTML types`, validates `hover, focus, and active visual states, `confirms the documented `minimum width and mobile full-width layout`, and verifies that clicking the rendered button dispatches the expected `click event.`

**Changes**
**`E2E tests`** (`ontario-button.e2e.ts`, replacing a fully-commented-out legacy file): hydration, aria-labels in both languages,  click events, hover and focus and @axe-core/playwright accessibility scans.

**VRT tests** (`ontario-button.vrt.ts`, new file):  These tests cover the button variants, interaction states, link rendering, and mobile layout across chromium, firefox, and webkit.


packages/ontario-design-system-component-library/src/components/ontario-button/test/ontario-button.vrt.ts
```diff
@@ -1,91 +1,107 @@
...
-test.describe('ontario-button - default states', () => {
-	test.beforeEach(async ({ page }) => {
-		await page.setContent(`
-			<ontario-button type="primary" element-id="ontario-button-primary">Primary</ontario-button>
-			<ontario-button type="secondary" element-id="ontario-button-secondary">Secondary</ontario-button>
-			<ontario-button type="tertiary" element-id="ontario-button-tertiary">Tertiary</ontario-button>
-		`);
-		await page.waitForChanges();
-	});
+const buttonTypes = ['primary', 'secondary', 'tertiary'] as const;
```
smorrisods: This re-declares a subset of `ButtonTypes` from `ontario-button.types.ts` (which exports `['primary', 'secondary', 'tertiary', 'internalThemeDark'] as const`) as a local literal array. Since we already use the single-source-of-truth pattern elsewhere (e.g. `ontario-badge.types.ts`), let's import `ButtonTypes` directly here instead — this VRT suite would then also cover `internalThemeDark` (which currently gets zero visual regression coverage), and any future button type addition gets baseline screenshots automatically instead of relying on someone remembering to update this array too.

packages/ontario-design-system-component-library/src/components/ontario-button/test/ontario-button.e2e.ts
```diff
@@ -1,81 +1,254 @@
...
+					button.addEventListener('click', () => resolve(), { once: true });
+				}),
+		);
+
+		await host.locator('button').click();
+		await clickPromise;
+	});
+
+	for (const buttonType of ['primary', 'secondary', 'tertiary'] as const) {
+		for (const language of [
```
smorrisods: Same pattern here — this re-declares a subset of `ButtonTypes`, which this very file already imports and iterates a few tests up (`for (const buttonType of ButtonTypes)`). Let's use `ButtonTypes` directly so `internalThemeDark` (already exercised for styles in `'applies the expected styles for each button type'`) gets an axe scan too, and this list can't quietly drift out of sync with the real button types.

packages/ontario-design-system-component-library/src/components/ontario-button/test/ontario-button.e2e.ts
```diff
@@ -1,81 +1,254 @@
...
+			{ name: 'English', label: 'Save profile' },
+			{ name: 'French', label: 'Enregistrer le profil' },
+		]) {
+			test(`the ${language.name} ${buttonType} button has no axe violations`, async ({ page }) => {
+				await page.setContent('<ontario-button></ontario-button>');
```
smorrisods: Tracing through `ontario-button.tsx`, there's no `language` prop and no `setAppLanguage`/`headerLanguageToggled` listener on this component — `label`/`ariaLabelText` are just plain strings. So `{ name: 'French', label: 'Enregistrer le profil' }` isn't exercising any French-language code path; it's really just passing a different string into the same prop. Naming this `` `the ${language.name} ${buttonType} button has no axe violations` `` implies we're validating bilingual support, when what's actually validated is that axe passes regardless of label content. I'd rename to something like `'has no axe violations with a non-ASCII label'` (and drop the language split down to one parametrized case), unless the plan is to eventually route this through a real language mechanism — worth a quick gut check either way.

packages/ontario-design-system-component-library/src/components/ontario-button/test/ontario-button.e2e.ts
```diff
@@ -1,81 +1,254 @@
...
+				await page.waitForChanges();
```
smorrisods: This block calls its own `page.setContent('<ontario-button></ontario-button>')` and re-queries `page.locator('ontario-button')`, duplicating what the file's `beforeEach`/shared `host` already provide for every other test in this suite. Let's use `host` here too for consistency, unless there's a specific reason this needs isolated setup.

smorrisods (changes requested):
Nice suite here, Tresa — the interaction-state and responsive coverage is thorough, and swapping to the shared `expectVrtScreenshot`/`withGlobalStyles` helpers is a good cleanup over the old per-test screenshot config. Found one recurring issue worth fixing before merge: both `ontario-button.e2e.ts` and `ontario-button.vrt.ts` re-declare a local `['primary', 'secondary', 'tertiary']` array instead of importing `ButtonTypes` from `ontario-button.types.ts`, which silently drops `internalThemeDark` from both the axe scan and VRT coverage. Also flagged a naming/intent question on the "French button" axe test and a minor setup-consistency nit. Details inline.

One scoping note: this branch is built on top of #364 (the blockquote e2e/VRT tests), which hasn't merged yet — the diff against `develop` shows `ontario-blockquote.e2e.ts`/`ontario-blockquote.vrt.ts` and the `HTMLOntarioBlockquoteElement` typing changes as part of this PR, but comparing this branch directly against #364's tip shows zero incremental changes to any blockquote file. In other words, all of that content is inherited unchanged from #364 rather than new work done here, so I've scoped this review to only the button-specific files (`ontario-button.e2e.ts`, `ontario-button.vrt.ts`, and the new button snapshots). Once #364 merges and this branch rebases onto `develop`, that diff noise should disappear on its own — no action needed here, just flagging it so the blockquote pieces aren't reviewed twice or mistaken for new work in this PR.

---
🤖 AI-drafted, Scott-flavoured. No warranty on the jokes, full warranty on the feedback.

packages/ontario-design-system-component-library/src/components/ontario-button/test/ontario-button.vrt.ts
tresajm: fixed

packages/ontario-design-system-component-library/src/components/ontario-button/test/ontario-button.e2e.ts
tresajm: fixed

packages/ontario-design-system-component-library/src/components/ontario-button/test/ontario-button.e2e.ts
tresajm: fixed

packages/ontario-design-system-component-library/src/components/ontario-button/test/ontario-button.e2e.ts
tresajm: fixed

PR #397, opened 2026-09-09
DS-1763: resolved React event prop naming mismatch

erincarrieresontario:
CX flagged that the docs for ontario-input listed the event as inputOnChange, but in React you actually need to use onInputOnChange — caused confusion/bugs for consumers wiring up the React components.

Turns out this same mismatch applies to all our components with custom events (ontario-input, ontario-textarea, ontario-date-input, ontario-search-box), since the Stencil React wrapper auto-generates prop names as on + capitalized event name (e.g. inputOnChange → onInputOnChange, searchOnSubmit → onSearchOnSubmit).

What changed:

- Added a note above each Events table clarifying the React prop naming convention
- Added full React code examples (alongside existing HTML/Angular ones) in the "Event model" sections so consumers can see the correct onXOnY prop usage directly, not just infer it
- No code/behavior changes — docs only


packages/ontario-design-system-component-library/src/components/ontario-textarea/readme.md
```diff
@@ -384,6 +491,8 @@ Disabled/read-only policy source:
 
 ## Events
 
+_Note: when using the React wrapper (`@ongov/ontario-design-system-component-library-react`), each event below is exposed as a prop named `on` + the event name with the first letter capitalized. For example, `inputOnChange` becomes `onInputOnChange`, and `inputOnInput` becomes `onInputOnInput`._
```
smorrisods, line 494: Let's not call it the react wrapper, we constantly present them as the React components (even though they're a wrapper internally).

I will also note, this is in an autogenerated block and will get wiped out when the auto generation happens.  We should call out the use of the name change in the React Component Lib readme (so it is published with the package) and then link to it if we want a per component reference.

smorrisods: Having the event based code samples across HTML, React, and Angular is a good idea that's used in this file.  This is a cleaner way to link back to the common description and it gives hands on examples.

packages/ontario-design-system-component-library/src/components/ontario-textarea/readme.md
```diff
@@ -209,6 +276,46 @@ document.querySelector('ontario-textarea').addEventListener('change', (event) =>
...
 
+```mdx-code-block
+</TabItem>
+<TabItem value="react">
+```
+
+React normalizes the native `input` event to `onChange`. Because the native
+`change` event does not cross the ShadowDOM boundary (see above), use
+`onInputOnChange` (mapped from the component's `inputOnChange` custom event)
+if you need to react to changes from React, rather than `onChange`.
```
smorrisods, line 287: Let's remove the hard line wrapping on this paragraph, it is out of place for our markdown files.

packages/ontario-design-system-component-library/src/components/ontario-textarea/readme.md
```diff
@@ -177,6 +190,47 @@ Eg. To access the value of any input entered into this component from the `input
...
+```mdx-code-block
+</TabItem>
+<TabItem value="react">
+```
+
+**Note:** in React, you don't attach these events with `addEventListener` like
+in the HTML example above. Instead, the React wrapper turns each custom event
+into its own prop, named by prefixing `on` and capitalizing the event name.
+So the `inputOnInput` custom event becomes the `onInputOnInput` prop, and
+`inputOnChange` becomes the `onInputOnChange` prop, as shown below.
```
smorrisods, line 202: Let's remove the hard line wrapping on this paragraph, it is out of place for our markdown files.

packages/ontario-design-system-component-library/src/components/ontario-search-box/readme.md
```diff
@@ -326,6 +377,8 @@ Disabled/read-only policy source:
 
 ## Events
 
+\_Note: when using the React wrapper (`@ongov/ontario-design-system-component-library-react`), each event below is exposed as a prop named `on` + the event name with the first letter capitalized. For example, `inputOnChange` becomes `onInputOnChange`, and `searchOnSubmit` becomes `onSearchOnSubmit`.
+
```
smorrisods, line 381: This will be auto-deleted when the auto-generation process runs in Stencil.

packages/ontario-design-system-component-library/src/components/ontario-search-box/readme.md
```diff
@@ -123,6 +136,44 @@ The following example registers a simple function on `window` `load` that adds a
...
 
+```mdx-code-block
+</TabItem>
+<TabItem value="react">
+```
+
+**Note:** in React, you don't attach this event with `addEventListener` like
+in the HTML example above. Instead, the React wrapper turns the
+`searchOnSubmit` custom event into its own prop, `onSearchOnSubmit`, as shown
+below.
```
smorrisods, line 147: Let's remove the hard line wrapping on this paragraph, it is out of place for our markdown files.

packages/ontario-design-system-component-library/src/components/ontario-input/readme.md
```diff
@@ -227,6 +296,48 @@ document.querySelector('ontario-input').addEventListener('change', (event) => {
...
 
+```mdx-code-block
+</TabItem>
+<TabItem value="react">
+```
+
+React normalizes the native `input` event to `onChange`. Because the native
+`change` event does not cross the ShadowDOM boundary (see above), use
+`onInputOnChange` (mapped from the component's `inputOnChange` custom event)
+if you need to react to changes from React, rather than `onChange`.
```
smorrisods, line 307: Let's remove the hard line wrapping on this paragraph, it is out of place for our markdown files.

packages/ontario-design-system-component-library/src/components/ontario-input/readme.md
```diff
@@ -514,6 +625,8 @@ Disabled/read-only policy source:
 
 ## Events
 
+_Note: when using the React wrapper (`@ongov/ontario-design-system-component-library-react`), each event below is exposed as a prop named `on` + the event name with the first letter capitalized. For example, `inputOnChange` becomes `onInputOnChange`, and `inputOnInput` becomes `onInputOnInput`. See the [Event model](#event-model) section above for React usage examples._
```
smorrisods, line 628: This will be auto-deleted when the auto-generation process runs in Stencil.

packages/ontario-design-system-component-library/src/components/ontario-input/readme.md
```diff
@@ -199,6 +212,49 @@ Example of the `inputOnInput` event:
...
+```mdx-code-block
+</TabItem>
+<TabItem value="react">
+```
+
+**Note:** in React, you don't attach these events with `addEventListener` like
+in the HTML example above. Instead, the React wrapper turns each custom event
+into its own prop, named by prefixing `on` and capitalizing the event name.
+So the `inputOnInput` custom event becomes the `onInputOnInput` prop, and
+`inputOnChange` becomes the `onInputOnChange` prop, as shown below.
```
smorrisods, line 224: Let's remove the hard line wrapping on this paragraph, it is out of place for our markdown files.

packages/ontario-design-system-component-library/src/components/ontario-date-input/readme.md
```diff
@@ -573,6 +632,8 @@ Disabled/read-only policy source:
 
 ## Events
 
+_Note: when using the React wrapper (`@ongov/ontario-design-system-component-library-react`), each event below is exposed as a prop named `on` + the event name with the first letter capitalized. For example, `inputOnChange` becomes `onInputOnChange`, and `inputOnInput` becomes `onInputOnInput`. See the [`onInputOnChange` example](/packages/app-nextjs/src/app/components/ontario-date-input/client-side/page.tsx) for reference._
```
smorrisods, line 635: This will be auto-deleted when the auto-generation process runs in Stencil.

packages/ontario-design-system-component-library/src/components/ontario-date-input/readme.md
```diff
@@ -368,6 +381,52 @@ Eg. To access the value of any change made to this component from the `inputOnIn
...
+```mdx-code-block
+</TabItem>
+<TabItem value="react">
+```
+
+**Note:** in React, you don't attach these events with `addEventListener` like
+in the HTML example above. Instead, the React wrapper turns each custom event
+into its own prop, named by prefixing `on` and capitalizing the event name.
+So the `inputOnInput` custom event becomes the `onInputOnInput` prop, and
+`inputOnChange` becomes the `onInputOnChange` prop, as shown below.
```
smorrisods, line 393: Let's remove the hard line wrapping on this paragraph, it is out of place for our markdown files.

smorrisods (changes requested):
Good find on calling out and adding samples for using the Events, especially in React where `on` is appended to each event.  I've added a few comments around this.  I think we need to add a section to the main React component library Readme then link to it so we don't scatter the same explanation everywhere.  I love the examples and we should flesh the same example concepts out across the component readmes for each framework.

TalhaAgro (commented):
+1 to Scott’s comments, if I understand correctly, the `ontario-search-box/readme.md` and `ontario-textarea/readme.md` files need to be regenerated with `pnpm build-libs` and then the resulting changes need to be committed

Otherwise LGTM!

Harris06V:
Agree with Scott and Talha's comments, nothing else to add

TalhaAgro (changes requested):

erincarrieresontario:
Removed the individual changes to the readmes and added the main guidance to the react-library readme

muntalee (approved):
LGTM @erincarrieresontario 🐐

